PR #250, opened 2015-10-07
remove clay from label of npm components

### LINKED ISSUE #243: Remove "Clay " from npm component labels
In the component labels, we're title-casing the names of the components. Now that we have a bunch of components on npm namespaced with `clay-*` the labels look a little weird.

Is it worth it to keep that "Clay " designation in the label, or is that extraneous? (I think it's extraneous)

yoshokatana:
fixes #243

### DIFF
--- a/services/label.js
+++ b/services/label.js
@@ -2,6 +2,16 @@ var _ = require('lodash'),
   references = require('./references');
 
 /**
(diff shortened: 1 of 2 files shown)


cruzanmo:
Nice, I like this solution.

+1

PR #258, opened 2015-10-27
first pass of responsive component selector

yoshokatana:
* new design
* menu opens/closes on click/tap
* no animations (yet)
* same functionality as before
* [trello ticket](https://trello.com/c/bUZtN6Pa/202-l-component-selector-update)

### DIFF
--- a/services/render.js
+++ b/services/render.js
@@ -42,8 +42,12 @@ function addComponentSelector(el) {
 
   return edit.getData(ref)
     .catch(function (e) {
-      console.log(ref)
-      console.error(e.message, e.stack);
+      // e.g. tried to get the schema of a component, and it 404'd
+      if (e.message === '404') {
+        return {}; // data is empty
+      } else {
+        return e; // something weird happened. don't eat the error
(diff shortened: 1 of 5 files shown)


services/render.js
```diff
@@ -41,6 +41,14 @@ function addComponentSelector(el) {
   var ref = el.getAttribute(references.referenceAttribute);
 
   return edit.getData(ref)
+    .catch(function (e) {
+      // e.g. tried to get the schema of a component, and it 404'd
```
yoshokatana: catching 404s earlier, so they don't get lumped in with other errors

TakenPilot: Why returning an object?  Why not null for the lack of an object?

TakenPilot:
Legit.  +1, though I have reservations about return an empty object.  That'll bite us later.

services/render.js
yoshokatana: fair point

TakenPilot:
That's more obvious, yeah.  +1

PR #260, opened 2015-10-29
allow soft-maxlength to calculate length when field is undefined

yoshokatana:
This, plus https://github.com/nymag/sites/pull/249 allow headlines to be edited again


gloddy:
:+1:

PR #265, opened 2015-11-05
UI Font Update

yoshokatana:
* use system fonts for all ui elements
* use lighter placeholder styles
* consolidate typgraphic styles
* fix line-breaks in inputs


yoshokatana:
Per @joshyporkchops we're using our current font stack, but consolidating it (beforehand we were using Arial and Helvetica interchangeably)

PR #269, opened 2015-11-09
input behavior

### LINKED ISSUE #267: consolidate text input behaviors
* `text`, `url`, etc should just be `input`
* add a `type` argument that accepts standard text inputs (e.g. `text`, `email`, `url`, `number`, `tel`, etc)
* throw error if `type` is non-text (e.g. `checkbox`, `radio`)

This will allow greater flexibility in inputs, make the code more DRY, and allow for more accurate input types for users.

yoshokatana:
* fixes #267
* consolidates `text` and `url` behaviors into `input`
* adds optional `type` argument for other text-like input types
* throws error on non-text-like `type` arguments (e.g. checkbox, button)


TakenPilot:
+1

PR #275, opened 2015-11-16
simplified component list

yoshokatana:
* reduced the logic to the simplest possible form
* client order overwrites server order, **always**
* so simple I can't even think of a test to write!


TakenPilot:
+1

PR #277, opened 2015-11-17
progress bars and status messages

yoshokatana:
* added `progress` service with tests
* `start()` and `done()` control progress bars
* `open()` and `close()` control status messages
* blue progress bar when saving
* green progress bar when publishing
* if there's a server error, progress bar turns red and finishes, then red status slides up
* if there's a validation error (when publishing), progress bar turns red and ends (usually too fast to see, but will be slower if we have a lot of validation rules) and validation error pane slides up
* yeah, I said "slides up." Panes slide up now.
* overlays fade in (very quickly) as well
* if you lose/gain connection, appropriate grey progress bars / status messages appear


styleguide/progress.scss
```diff
@@ -0,0 +1,46 @@
+@import 'colors';
+@import 'typography';
+
+// progress bars
+.kiln-progress-wrapper {
+  bottom: 100%;
```
cruzanmo: Why not `top: 0;`?

yoshokatana: bottom: 100% pushes it above the enclosing element, rather than just sitting at the top :-)

client.js
```diff
@@ -45,3 +47,22 @@ document.addEventListener('DOMContentLoaded', function () {
...
+});
+
+window.addEventListener('online', function () {
+  progress.close(); // in case there are any status messages open, close them
+});
+
+window.addEventListener('offline', function () {
+  progress.done('grey'); // turn any progress indicators to grey and end them
+  progress.open('grey', connectionLostMessage);
+});
```
TakenPilot: What if the visual changes for this?  You're going to search for all reference to "grey"?

This event shouldn't reference the visual, but should represent what happened so people listening to the event can respond to a thing that has occurred.  That is, it's weird that people are responding to a "code grey".

"Alert!  Alert!  We've gone to Code Fuchsia!"  "What does that mean?"  "I don't know!!"

styleguide/progress.scss
cruzanmo: Cool.

controllers/publish-pane.js
```diff
@@ -14,14 +15,24 @@ module.exports = function () {
     },
 
     onPublishNow: function () {
+      pane.close();
+      progress.start('green');
+
       return validation.validate(rules).then(function (errors) {
         if (errors.length) {
-          pane.close();
+          progress.done('red');
```
TakenPilot: Code red!  Hey, progress bar, be red now.

Really, shouldn't the progress bar be responsible for how it looks?  Making the publish pane responsible for how _someone else_ looks is an anti-pattern.

controllers/publish-pane.js
```diff
@@ -14,14 +15,24 @@ module.exports = function () {
...
           pane.openValidationErrors(errors);
         } else {
-          return edit.publishPage().then(function () {
-            alert('published!'); // eslint-disable-line
-          });
+          return edit.publishPage()
+            .then(function (url) {
+              progress.done();
+              progress.open('green', `Published! <a href="${url}" target="_blank">View Article</a>`);
```
TakenPilot: And now we're even filling in it's text for it.  :/

controllers/publish-pane.js
```diff
@@ -14,14 +15,24 @@ module.exports = function () {
...
+            })
+            .catch(function () {
+              // note: the Error passed into this doesn't have a message, so we use a custom one
+              progress.done('red');
+              progress.open('red', `A server error occured. Please try again.`, 3500);
```
TakenPilot: Magic number.

services/edit/index.js
```diff
@@ -95,6 +96,7 @@ function save(data) {
   var uri = data[refProp],
     schemaPromise = data._schema && Promise.resolve(data._schema) || cache.getSchema(uri);
 
+  progress.start('blue');
```
TakenPilot: "Start Blue!"

"Beginning Blue."

"Blue has begun."

client.js
yoshokatana: Agreed, that's what I was thinking about. What sounds better:

1. Passing in a "semantic" word like `progress.start('publish')`
2. Export the colors object and reference it directly, like `progress.start(progress.colors.publish)`

styleguide/pane.scss
```diff
@@ -41,6 +43,11 @@ $pane-margin: 30vh;
   @media screen and (min-width: 1024px) {
     width: 80%;
   }
```
TakenPilot: If you use a percentage width, you should usually combine it with a min-width.

styleguide/progress.scss
```diff
@@ -0,0 +1,46 @@
...
+  left: 0;
+  pointer-events: none;
+  position: absolute;
+  width: 100%;
```
TakenPilot: width: 100% vs display block...

You're using width: 100% a lot.  Why?

styleguide/progress.scss
```diff
@@ -0,0 +1,46 @@
...
+    position: absolute;
+    width: 100%;
+  }
+}
+
+// status messages
+.kiln-status {
+  @include primary-text();
+
+  bottom: 100%;
```
TakenPilot: bottom: 100%?

bottom: 0 ?

client.js
TakenPilot: Things should be responsible to drawing themselves.

Start('publish') would make more sense.

styleguide/progress.scss
TakenPilot: Are you growing outside of your own space?  This is like negative margins.  Very dangerous, yeah?

styleguide/progress.scss
TakenPilot: And therefore not defending your space.  Using negative space like this means adding anything around your DOM element is going to be very difficult in the future.

yoshokatana: it's absolutely positioned, and above everything else (or rather, it lives in the same place as the toolbar)

styleguide/pane.scss
yoshokatana: good catch, added.

styleguide/progress.scss
yoshokatana: I...want it to be 100% width

styleguide/progress.scss
yoshokatana: In most circumstances, yes. In this one, it's overlaying other elements on the page (and it doesn't care)

yoshokatana:
Per Dane:

> Man, push through what you want, but I warned ya.  You're going to be returning to this code every week or so, or basically whenever you do anything ​_around_​ it, because your API is fragile.

PR #279, opened 2015-11-23
block repeated items in simple-list

yoshokatana:
* to repeat items, pass in the `allowRepeatedItems` argument
* this uses native `setCustomValidity` to make the input invalid until you try and add a different tag
* [trello ticket](https://trello.com/c/UISBXyTB/198-no-repeated-items-in-simple-lists)
* also added arrow functions to eslintrc
* NOTE: tags are case insensitive for purposes of comparison


ughitsaaron:
:+1:

PR #282, opened 2015-11-30
block form closing if it fails validation

yoshokatana:
* [trello ticket that prompted this](https://trello.com/c/3wPlH7Bm/55-bug-image-and-video-components-should-not-be-able-to-be-published-with-invalid-urls)
* forms will not close if they have blocking (e.g. native validation) errors
* trying to close form will trigger native validation messages
* other services that try and close forms will hit the validation as well

### DIFF
--- a/controllers/overlay.js
+++ b/controllers/overlay.js
@@ -42,7 +42,7 @@ module.exports = function () {
     close: function (e) {
       if (e.target === e.currentTarget) {
         // we clicked on the overlay itself
-        focus.unfocus();
+        focus.unfocus().catch(_.noop);
       }
     }
   };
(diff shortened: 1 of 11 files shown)


cruzanmo:
`Firefox 42.0.0 (Mac OS X 10.10.0) ERROR`

controllers/overlay.js
```diff
@@ -42,7 +42,7 @@ module.exports = function () {
     close: function (e) {
       if (e.target === e.currentTarget) {
         // we clicked on the overlay itself
-        focus.unfocus();
+        focus.unfocus().catch(_.noop);
```
cruzanmo: This allows the form to close when it is invalid?

yoshokatana: No, that's a noop. It'll stay open, but not throw an error

cruzanmo: Oh, I see, that makes sense.

cruzanmo:
+1

PR #285, opened 2015-12-03
kiln stability improvements

yoshokatana:
* confirm before creating a new page (if you have forms open)
* save/close forms before opening publish pane (like we did before)
* POST page _url_ instead of page _uri_ to `/schedule`
* look for already-published url in `page@published`, rather than just `page`
* added `db.uriToUrl` to match `db.urlToUri`


TakenPilot:
+1

PR #293, opened 2015-12-07
fix issue with iframe overlays

yoshokatana:
[trello ticket](https://trello.com/c/PMkdkCTp/173-bugs-firefox-specific)


TakenPilot:
Decreasing coverage...  okay.  +1

yoshokatana:
It's decreasing, but is less than whatever the cutoff is (0.1% I think)? So it's ok. :p

PR #295, opened 2015-12-08
fixes for mobile editing, scrolling, zooming, and tapping

yoshokatana:
* set font size to `16px` to prevent mobile zooming (it's also easier to read with my tired old eyes)
* change scrolling styles to (hopefully) prevent forced scrollbars in certain situations
* force offloading of certain fixed position stuff onto the gpu
* fix selection and form closing event handling (so we can close forms ALWAYS, even when selecting components)
* prevent some ios safari weirdness when focusing elements

### DIFF
--- a/controllers/form.js
+++ b/controllers/form.js
@@ -44,12 +44,18 @@ module.exports = function () {
   constructor.prototype = {
     events: {
       submit: 'closeForm',
-      close: 'closeForm'
+      close: 'closeForm',
+      'input focusin': 'stopFocus',
+      'input focus': 'stopFocus'
     },
 
     closeForm: function (e) {
       e.preventDefault();
       focus.unfocus().catch(_.noop);
+    },
+
+    stopFocus: function (e) {
+      e.preventDefault();
     }
   };
 
--- a/controllers/form.js
+++ b/controllers/form.js
@@ -46,7 +46,9 @@ module.exports = function () {
       submit: 'closeForm',
       close: 'closeForm',
       'input focusin': 'stopFocus',
-      'input focus': 'stopFocus'
+      'input focus': 'stopFocus',
+      '[contenteditable]': 'stopFocus',
+      'select': 'stopFocus'
     },
 
     closeForm: function (e) {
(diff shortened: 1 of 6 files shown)


controllers/form.js
```diff
@@ -44,12 +44,20 @@ module.exports = function () {
...
+      'select': 'stopFocus'
     },
 
     closeForm: function (e) {
       e.preventDefault();
       focus.unfocus().catch(_.noop);
+    },
+
+    stopFocus: function (e) {
+      e.preventDefault();
```
yoshokatana: this prevents mobile browsers from doing weird resizing things (separate from the zooming issue). Focusing actually works fine, and you can scroll through inputs as you'd expect (with native browser arrows, or by tapping other inputs).

styleguide/_typography.scss
```diff
@@ -16,10 +16,10 @@ $system-fonts: Helvetica, Arial, sans-serif;
 // our typography
 @mixin kiln-copy {
   font-family: $system-fonts;
-  font-size: 14px;
+  font-size: 16px;
```
yoshokatana: this prevents mobile safari (and supposedly other mobile browsers) from zooming in the page and breaking all our styles. also it looks better.

styleguide/overlay.scss
```diff
@@ -11,9 +11,11 @@ $overlay-margin: 10vh;
   height: 100%;
   left: 0;
   opacity: 0;
-  overflow: scroll;
+  overflow: auto;
```
yoshokatana: overflow: scroll isn't needed anymore for mobile scrolling, but _does_ force scrollbars in certain situations. auto + `-webkit-overflow-scrolling: touch` is the current best practice for making momentum scrolling happen.

styleguide/overlay.scss
```diff
@@ -11,9 +11,11 @@ $overlay-margin: 10vh;
...
+  -webkit-overflow-scrolling: touch;
   position: fixed;
   top: 0;
+  transform: translateZ(0); // create a new viewport context
```
yoshokatana: This will actually force most browsers to offload more to the GPU, and will create a new "viewport context". Normally creating a new viewport context is a _weird, unexpected behavior_ but it's fine when doing it on a fixed position element.

PR #305, opened 2015-12-10
add remove formatting button

yoshokatana:
Adds a "remove formatting" button to the end of the wysiwyg toolbar buttons, which allows users to quickly get rid of formatting in their text.

This is very useful if users are copy+pasting from sources that aren't strict about their formatting (CQ, Microsoft Word, etc) and include things like `<span style="custom style stuff">`. Without this button, it's impossible to get rid of those spans in the kiln UI.


TakenPilot:
+1

PR #309, opened 2015-12-15
clone components inside a component list when cloning the parent component

yoshokatana:
[trello ticket](https://trello.com/c/Kh8swqKX/32-related-stories-new-style)

When you add components via the <kbd>add component</kbd> button, it calls `edit.createComponent` which clones the base data. If that base data contains a component list, this will now clone those components as well.

For example, say I have a component called "Related Stories", that looks like this:

```yaml
components:
  related-stories:
    content:
      -
        _ref: /components/related-story
```

Now that `related-story` will be cloned, rather than having a direct reference to the base data.

### DIFF
--- a/services/edit/index.js
+++ b/services/edit/index.js
@@ -214,6 +214,54 @@ function createPage() {
 }
 
 /**
+ * create a child component (from a base ref) and return it's newly generated ref
+ * @param {string} baseRef e.g. [siteprefix]/components/name
+ * @returns {Promise}
+ */
+function createChildComponent(baseRef) {
+  return cache.getDataOnly(baseRef).then(function (baseData) {
+    return cache.createThrough(baseRef + '/instances', baseData).then(function (res) {
+      return res._ref;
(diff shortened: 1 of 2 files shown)


cruzanmo:
+1

services/edit/index.js
```diff
@@ -214,6 +214,54 @@ function createPage() {
...
+      // assumes a component will have a maximum of one component list
+    }),
+    items = componentList && res[componentList],
+    promises = items && _.map(items, function (item) {
+      return createChildComponent(item._ref);
+    });
+
+  if (items && items.length) {
+    return Promise.all(promises).then(function (newRefs) {
+      var newRes = _.cloneDeep(res); // create a new object, since we're explicitly modifying the parent component data
```
cruzanmo: What happens if `res` is updated in-place (without cloning)?

yoshokatana: it's read-only :-)

PR #314, opened 2015-12-17
only show add-component buttons for selected and parent-selected components

### LINKED ISSUE #271: rethink add component button
It's broken when things aren't 100% width, vertically-stacked:

* when components are floated
* when they're inline (or potentially table-stuff)
* when they're flexed

yoshokatana:
This will only show the add component buttons / menus for selected components, and for their immediate parent. Some examples:

* You select related-story. You see the plus button for related-stories
* You seelect related-stories. You see the plus button for it and its parent, article
* You select paragraph. You see the plus button for article

This fixes #271 (for the most part, though weird reflows will still happen. They won't be as big a deal now) and has a [trello ticket](https://trello.com/c/IH2kF3Rd/245-plus-buttons-live-inside-parent-component)

### DIFF
--- a/decorators/component-list.js
+++ b/decorators/component-list.js
@@ -253,11 +253,12 @@ function handler(el, options) {
   button = dom.find(pane, '.open-add-components');
 
   // add click events to toggle pane
-  button.addEventListener('click', function () {
+  button.addEventListener('click', function (e) {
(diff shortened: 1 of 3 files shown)


cruzanmo:
+1

PR #357, opened 2016-01-25
fix #344

### LINKED ISSUE #344: hit ESC to close forms
Users should be able to close _any_ form they're in by hitting <kbd>esc</kbd>. This will have the exact same behavior in overlays, settings forms, inline forms, and `wysiwyg` fields with `enableKeyboardExtras` toggled.

yoshokatana:
fixes #344

pressing <kbd>esc</kbd> will ALWAYS call `focus.unfocus()`, closing forms and unselecting components.

Note: if you have a form that is invalid, it will not close

### DIFF
--- a/controllers/kiln-toolbar.js
+++ b/controllers/kiln-toolbar.js
@@ -47,6 +47,16 @@ EditorToolbar = function (el) {
     }
   });
 
+  // close ANY open forms if user hits ESC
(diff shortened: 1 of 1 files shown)


TakenPilot:
Thank you! +1

(Link to issue?)

yoshokatana:
Right! Added.

PR #361, opened 2016-01-25
smoothly scroll to parent

### LINKED ISSUE #348: transition when scrolling to parent component
Right now clicking the parent component in a component selector is a jarring jump. We should have some kind of animation/slow scroll to it, to make it easier for users to visualize the hierarchy.

Note: Make sure you don't mess up the user scrolling if they try to manually scroll during this transition.

yoshokatana:
fixes #348

* smoothly animates scrolling when clicking parent selector
* stops scrolling if user initiates a manual scroll (on ff, chrome, safari, ios, etc)

### DIFF
--- a/services/scroll.js
+++ b/services/scroll.js
@@ -1,3 +1,39 @@
+const easingEquations = {
+  easeOutSine: function (pos) {
+    return Math.sin(pos * (Math.PI / 2));
+  },
+  easeInOutSine: function (pos) {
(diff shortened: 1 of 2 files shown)


yoshokatana:
I tried my hand at writing some tests against `scroll.js`, but they're super low ROI. I think we're fine with the slightly decreased coverage.

PR #374, opened 2016-02-08
use watchify instead of gulp.watch for scripts

yoshokatana:
* this allows browserify to cache script bundling
* decreases subsequent script compilation times (6s down to 100ms)
* pass `--once` to not compile once (with uglify)
* by default, `gulp scripts` will watch and WON'T uglify the js
* removed shoddy sourcemap support

Note: When running `npm postinstall` scripts, it'll compile once.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -56,7 +56,7 @@ gulp.task('scripts', function () {
       cache: {},
       packageCache: {}
     }),
-    runOnce = process.argv.indexOf('--once') > -1;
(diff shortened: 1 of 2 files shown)


TakenPilot:
Could we change this to be the same as npm?  Like, with a `--production` flag instead?  Is the `--once` stuff not us?

yoshokatana:
yeah, that works

TakenPilot:
+1

PR #384, opened 2016-02-24
set caret after pasted html

### LINKED ISSUE #366: wysiwyg: caret not being set at end when pasting text with formatting
In the `wysiwyg` behavior, if you paste plain text it sets the caret at the end. :cool: 

If the text contains elements (like `<em>` or `<a>`), it'll set the caret right before the first inline element.

It should set the caret at the end.

yoshokatana:
fixes #366

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -440,7 +440,7 @@ function initWysiwygBinder(enableKeyboardExtras) {
         // SelectionAPI R'lyeh fhtagn
         range = document.createRange();
         selection = window.getSelection();
-        range.setStart(editable, 1);
+        range.setStartAfter(editable.lastChild);
         range.collapse(true);
         selection.removeAllRanges();
         selection.addRange(range);


yoshokatana:
I also tested that this works fine when pasting text without inline elements, like normal.

behaviors/wysiwyg.js
```diff
@@ -440,7 +440,7 @@ function initWysiwygBinder(enableKeyboardExtras) {
         // SelectionAPI R'lyeh fhtagn
         range = document.createRange();
         selection = window.getSelection();
-        range.setStart(editable, 1);
+        range.setStartAfter(editable.lastChild);
```
TakenPilot: Does this have to be an element?  There are other kinds of DOM nodes.

There is code in dom.js to get the lastElement rather than last child, or something like that, somewhere.

cruzanmo:
nice +1

behaviors/wysiwyg.js
cruzanmo: Looks like setStartAfter takes Node not Element https://developer.mozilla.org/en-US/docs/Web/API/Range/setStartAfter

yoshokatana: `Node#lastChild` returns the last child (not just the last element) of a node. This is what we want :-)

PR #389, opened 2016-03-03
adds fake bulleted lists

yoshokatana:
[trello ticket](https://trello.com/c/mnKmye4Z/11-l-paragraph-editing)

This adds bullets when hitting <kbd>tab</kbd>. No, they aren't `<ul><li>` unordered lists, and no they aren't semantic. But on the other hand, they work perfectly and have extremely simple logic.

Let's use these until we get prosemirror.

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -345,10 +345,17 @@ function addStyledClass(styled) {
 }
 
 /**
+ * Add a bullet (for fake bulleted lists) and set the caret after it
+ */
+function addBullet() {
+  document.execCommand('insertHTML', false, '&bull;&nbsp;');
+}
+
+/**
  * Add a line break and set the caret after it
  */
 function addLineBreak() {
-  document.execCommand('insertHTML', false, '<br>');
+  document.execCommand('insertHTML', false, '<br><br>');
 }
 
 /**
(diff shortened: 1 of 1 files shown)


yoshokatana:
By the by +1 from christian and morgan

TakenPilot:
?!

behaviors/wysiwyg.js
```diff
@@ -345,10 +345,17 @@ function addStyledClass(styled) {
...
+function addBullet() {
+  document.execCommand('insertHTML', false, '&bull;&nbsp;');
+}
+
+/**
  * Add a line break and set the caret after it
  */
 function addLineBreak() {
-  document.execCommand('insertHTML', false, '<br>');
+  document.execCommand('insertHTML', false, '<br><br>');
```
yoshokatana: this fixes inserting line breaks when the caret is at the end of the line

yoshokatana:
And on the pull request these words appear:
'My name is Ozymandias, king of kings:
Look on my works, ye Mighty, and despair!'
Nothing semantic remains.

yoshokatana:
pullrequest

yoshokatana:
> “Semiotics is in principle the discipline studying everything which can be used in order to lie. If something cannot be used to tell a lie, conversely it cannot be used to tell the truth: it cannot in fact be used 'to tell' at all.”

yoshokatana:
Ideas:

* this paen to the elder gods
* something in paragraph settings where you type a bunch of hyphens and it converts on save
* some kind of separate "bulleted list" component, but I can't figure out how that would work

Trying to do "real" bulleted lists inside paragraphs yields... interesting results.
screen shot 2016-03-03 at 5 20 56 pm

PR #392, opened 2016-03-09
Publish layouts

yoshokatana:
* <kbd>publish now</kbd> publishes page and layout
* <kbd>schedule publish</kbd> schedules publishing of page and layout
* <kbd>unschedule</kbd> unschedules page _and_ layout publishing (note to keep in mind: if you schedule a layout publish, then unschedule on a different page that's also scheduled, the layout will be unscheduled)
* <kbd>unpublish</kbd> will only make the _page_ private. it will not revert any layout changes

Changelog

* added pub/schedule/unschedule functionality
* added messaging to publish pane
* NOTE: non-article pages _won't_ display as published anymore (so you won't get the status message or be able to unpublish them). This is due to us checking the article date in page-status. In the _near_ future we'll have proper timestamps to work with and this will work again

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -45,7 +45,7 @@
     "vars-on-top": 2,
     "wrap-iife": 2,
     // strict mode
-    "strict": [2, "safe"],
+    "strict": [2, "never"],
     // variables
     "no-unused-vars": 2,
     // node.js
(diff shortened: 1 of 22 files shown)


TakenPilot:
+1

PR #393, opened 2016-03-10
dynamic placeholder heights

yoshokatana:
* uses declared placeholder height, falls back to 100px (as before)
* if parent component has a set height (and that height is larger than the declared placeholder height), placeholder will fill vertical space
* this makes ads (and other components where instances may have different explicitly-set heights) look really nice in edit mode :sparkles:


TakenPilot:
+1

PR #403, opened 2016-03-31
added rollupify

yoshokatana:
* use rollupify in `gulp scripts`
* use rollupify in `karma`
* updated `cid` service to test it out

NOTE: As we migrate modules to the es6 module syntax, make sure to do modules with the fewest dependencies first.


yoshokatana:
:+1: from byron

PR #407, opened 2016-04-04
magic button

yoshokatana:
This adds a new behavior I'm calling :sparkles: magic button :sparkles: 

In brief, it grabs data from a field, optionally does stuff to it (including transforms, api calls, etc) and puts it into the current field.

All arguments to `magic-button` are optional.

* if you specify `field`, it'll grab the data from that field (otherwise it'll use emptystring)
* if you specify `transform` it'll run that data through a transform function (available transforms live in the behavior)
* if you specify `url`, it'll do a GET request against that `url + transformed data` (useful for mediaplay images!)
* if you specify `property`, it'll grab that property from the returned api call

Note: It **won't** update the current field if the data is empty, which prevents it from overwriting good data with garbage (in most situations).

Note: It'll **always** grab the same data (unless you do an api call that returns different data), so the behavior is super consistent.

Note: You can always change what's in the field after you press the button, or even erase the data entirely! There's no weird business logic going on.

**It can make headlines easier!**

**It can make image captions and credits easier!**


behaviors/magic-button.js
```diff
@@ -0,0 +1,168 @@
...
+
+      // remove rendition stuff
+      let barePath = path.replace(/\.w\d+/, ''); // remove width
+
+      barePath = barePath.replace(/\.h\d+/, ''); // remove height
+      barePath = barePath.replace('.2x', ''); // remove resolution
+
+      return barePath;
+    }
+  };
```
cruzanmo: Wondering how this will grow. Should the transformer be passed into the behavior?

yoshokatana: I didn't want to specify actual (runnable) code in the schema, and I think this'll remain pretty small.

behaviors/magic-button.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  if (!_.isEmpty(url)) {
+    // do an api call!
+    return getAPI(url + transformed)
+      .then(getProperty(property))
+      .then(res => setFieldData(bindings, currentField, res));
+  } else {
+    // just set the data
+    return setFieldData(bindings, currentField, transformed);
```
cruzanmo: Does not return a Promise

yoshokatana: whoops, updated

behaviors/magic-button.scss
```diff
@@ -0,0 +1,29 @@
...
+  top: 0; // allow 1px for border
+  width: 44px; // explicit width, see below
+
+
+  img {
+    height: 22px;
+    width: 24px;
+  }
+
+  & + * {
```
cruzanmo: Nice

cruzanmo:
+1

PR #412, opened 2016-04-07
eliminate extra api call when saving components

yoshokatana:
This eliminates the extra api call (to get the new rendered template) when saving components. It makes these things faster and more fluid:

* exiting out of component forms
* merging (deleting) paragraphs

It also makes the component creation api calls run in parallel, so those will be marginally faster (we can eliminate the second call entirely when we can POST to `/components/name/instances.html`)

☆.。.:*・°☆.。.:*・°☆.。.:*・°☆.。.:*・°☆
——— **2x speed improvements** ———
☆.。.:*・°☆.。.:*・°☆.。.:*・°☆.。.:*・°☆

### DIFF
--- a/services/edit/db.js
+++ b/services/edit/db.js
@@ -298,6 +298,22 @@ function save(uri, data) {
 }
 
 /**
+ * save, expecting html to be returned from the server
+ * @param {string} uri
+ * @param {object} data
+ * @returns {Promise}
+ */
+function saveForHTML(uri, data) {
+  assertUri(uri);
+
+  return send(addJsonHeader({
+    method: 'PUT',
+    url: uri + extHtml + '?edit=true',
+    data: data
+  })).then(expectHTMLResult(uri));
+}
+
+/**
  * @param {string} uri
  * @param {object} data
  * @returns {Promise}
@@ -361,6 +377,7 @@ module.exports.getText = getText;
 module.exports.getHead = getHead;
 module.exports.getHTML = getHTML;
 module.exports.save = save;
+module.exports.saveForHTML = saveForHTML;
 module.exports.saveText = saveText;
 module.exports.create = create;
 module.exports.remove = remove;
--- a/services/edit/db.js
+++ b/services/edit/db.js
@@ -4,6 +4,9 @@ var _ = require('lodash'),
   references = require('../references'),
   site = require('./../site'),
   extHtml = '.html',
+  editMode = '?edit=true',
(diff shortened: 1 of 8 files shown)


services/edit/db.js
```diff
@@ -298,6 +298,22 @@ function save(uri, data) {
...
+ * @param {string} uri
+ * @param {object} data
+ * @returns {Promise}
+ */
+function saveForHTML(uri, data) {
+  assertUri(uri);
+
+  return send(addJsonHeader({
+    method: 'PUT',
+    url: uri + extHtml + '?edit=true',
```
cruzanmo: Could store `'?edit=true'` at top like `extHtml`

yoshokatana: Good idea

services/render.js
```diff
@@ -81,12 +81,13 @@ function addComponentsHandlers(el) {
...
-function reloadComponent(ref) {
-  return db.getHTML(ref)
-  // ?edit=true so that components will return back edit-mode stuff (if they check for locals.edit)
-  // e.g. clay-tweet and clay-facebook-post load scripts in view mode, but not edit mode
+function reloadComponent(ref, html) {
+  var serverHtml = html ? Promise.resolve(html) : db.getHTML(ref);
+
+  return serverHtml
     .then(function (el) {
       var currentEls = dom.findAll('[' + references.referenceAttribute + '="' + ref + '"]');
```
ughitsaaron: could use template literal here :information_desk_person:

services/edit/index.js
```diff
@@ -363,8 +364,15 @@ function addToParentList(opts) {
...
-    return save(parentData)
-      .then(db.getHTML.bind(null, ref));
+    return Promise.all([
+      // save the parent and get the child's html in parallel
+      // note: this assumes the child already exists
+      // todo: when we can POST and get html back, just handle the parent here
+      save(parentData),
+      db.getHTML(ref)
+    ]).then(function (results) {
+      return results[1]; // return the child component's html
```
ughitsaaron: `results => results[1]`?

services/edit/db.js
yoshokatana: done

services/edit/index.js
yoshokatana: done

ughitsaaron:
:+1:

yoshokatana:
gotta chase that 0.006%!

PR #416, opened 2016-04-11
refactored component creation

yoshokatana:
* when components are created, it checks to see if there are child components that also need to be created
* (this check is more robust than it was before)
* it generates a mapping object, creates the parent and children in parallel, then applies the child instance IDs to the parent (and updates it)

### DIFF
--- a/services/edit/cache.js
+++ b/services/edit/cache.js
@@ -100,10 +100,36 @@ function getSchema(uri) {
 }
 
 /**
+ * save component, get saved data back
  * @param {object} data
  * @returns {Promise}
  */
 function saveThrough(data) {
+  var uri = data[references.referenceProperty];
+
+  return removeExtras(uri, data).then(function (data) {
+    return db.save(uri, data);
+  }).then(function (result) {
+    // only clear cache if save is successful
+    exports.getData.cache = new _.memoize.Cache();
+    exports.getDataOnly.cache = new _.memoize.Cache();
+
+    // remember new value (it is returned when save is successful; common with REST implementations)
+    result[references.referenceProperty] = uri;
+    control.setReadOnly(result);
+    exports.getDataOnly.cache.set(uri, result);
+
+    // cache version with schema, return version with schema
+    return exports.getData(uri);
+  });
+}
+
+/**
+ * save component, get html back
+ * @param {object} data
+ * @returns {Promise}
+ */
+function saveForHTML(data) {
   var uri = data[references.referenceProperty];
 
   return removeExtras(uri, data).then(function (data) {
@@ -148,4 +174,5 @@ exports.getSchema = control.memoizePromise(getSchema);
 
 // forgets
 exports.saveThrough = saveThrough;
+exports.saveForHTML = saveForHTML;
 exports.createThrough = createThrough;
--- a/services/edit/cache.test.js
+++ b/services/edit/cache.test.js
@@ -111,6 +111,62 @@ describe('cache service', function () {
       var data = {_ref: 'foo'};
 
       db.get.returns(Promise.resolve({foo: 'bar'}));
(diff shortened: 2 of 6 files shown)


services/edit/cache.js
```diff
@@ -100,13 +100,39 @@ function getSchema(uri) {
...
+    return exports.getData(uri);
+  });
+}
+
+/**
+ * save component, get html back
+ * @param {object} data
+ * @returns {Promise}
+ */
+function saveForHTML(data) {
```
yoshokatana: I brought `saveThrough` back (since it's still useful), and moved the "save but gimme html plz" logic to a new function, `saveForHTML`

services/edit/cache.js
```diff
@@ -148,4 +174,5 @@ exports.getSchema = control.memoizePromise(getSchema);
 
 // forgets
 exports.saveThrough = saveThrough;
+exports.saveForHTML = saveForHTML;
 exports.createThrough = createThrough;
```
yoshokatana: eventually (when we have a "create but gimme html plz" endpoint), we could also have a `createForHTML` function

services/edit/cache.test.js
```diff
@@ -111,6 +111,62 @@ describe('cache service', function () {
...
+      db.get.returns(Promise.resolve(fetched));
+      db.save.returns(Promise.resolve(returnSelf));
+      db.getSchema.returns(Promise.resolve(schema));
+      return fn(data).then(function () {
+        expect(lib.getData.cache.get(uri)).to.deep.equal({foo: 'bar2', _ref: uri, _schema: schema});
+      });
+    });
+  });
+
+  describe('saveForHTML', function () {
```
yoshokatana: I moved the `saveThrough` tests to `saveForHTML` and brought in the old `saveThrough` tests

services/edit/index.js
```diff
@@ -292,16 +328,26 @@ function cloneChildComponents(res) {
...
-    return cache.getDataOnly(base) // create component with base JSON from bootstrap.
-      .then(function (baseJson) {
-        return cache.createThrough(instance, baseJson).then(cloneChildComponents);
-      });
+    // otherwise get the component's default data
+    getDefaultData = cache.getDataOnly(base);
   }
+
+  return getDefaultData.then(function (defaultData) {
+    let children = getChildComponents(defaultData);
```
yoshokatana: The _major_ philosophical difference with this change is that we're figuring out the child components and creating them **in parallel**, rather than waiting until we've already created the parent before looking for children.

services/promises.js
```diff
@@ -19,6 +19,31 @@ function defer() {
 }
 
 /**
+ * bluebird.props with native promises
+ * @param {object} props
+ * @returns {Promise}
+ */
+function props(props) {
```
yoshokatana: This works the same as `bluebird.props`, but I didn't want to pull in the _whole_ bluebird lib just for one function.

byronhulcher:
This looks good to me, might be worthwhile to have another person +1

services/edit/index.js
```diff
@@ -234,51 +235,85 @@ function createPage() {
...
-      return res._ref;
-    });
+function getChildComponents(data) {
+  let mapping = {};
+
+  _.forOwn(data, function checkProperties(val, key) {
+    // loop through the (base) data in the component
+    if (_.isArray(val)) {
+      // find arrays
+      _.each(val, function checkArrays(item, index) {
```
larrychevres: Is there a functional or style reason for using named function on the forOwn and each iterations?

larrychevres:
+1, one minor question noted in code

services/edit/index.js
yoshokatana: Nope, it just makes for easier debugging via stack trace if anything goes wrong

PR #418, opened 2016-04-13
remove input dep from wysiwyg

### LINKED ISSUE #268: remove input dependency from wysiwyg behavior
Right now, you have to apply the `wysiwyg` behavior on top of something like `text` or `textarea`. It should just be a first-class behavior, since there are no differences in styling or functionality.

### LINKED ISSUE #346: make soft-maxlength clearer
Right now `soft-maxlength` displays the remaining characters like `Remaining: 20`. We should change that to also show the total characters allowed, to give users more context. Options:

* `20/100`
* `Remaining: 20/100`
* ???

yoshokatana:
* remove input dep from wysiwyg (fixes #268)
* wysiwyg input should inherit `word-break` property when unstyled
* show "max" length in `soft-maxlength` (fixes #346)

### DIFF
--- a/behaviors/soft-maxlength.js
+++ b/behaviors/soft-maxlength.js
@@ -31,24 +31,19 @@ function setText(text, span) {
 /**
  * set styles depending on the remaining length
  * @param {number} remaining
+ * @param {number} max
  * @param {Element} el
  */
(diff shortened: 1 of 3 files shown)


ughitsaaron:
nice

PR #422, opened 2016-04-14
treat simple-list like inputs

yoshokatana:
treats simple lists like inputs:

* wraps them in `.input-label` so labels can be applied
* updates `getInput` to match them so other behaviors can be added before / after

screen shot 2016-04-14 at 11 13 32 am


yoshokatana:
:+1: from byron

PR #423, opened 2016-04-14
paste multiple paragraphs

yoshokatana:
* paste multiple paragraphs
* if there's already text and you paste a single paragraph, it'll append to the text
* if there are more paragraphs, it'll create them
* it runs all pasted text through `text-model`
* paragraphs are created in parallel, for speed (⌐■_■)
* after paragraphs are created, it focuses on the last one and sets the caret to the end
* it's not instant, but it's pretty fast

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -128,7 +128,8 @@ function createEditor(field, buttons) {
       preCleanReplacements: [
         [/<h[1-9]>/ig, '<h2>'],
         [/<\/h[1-9]>/ig, '</h2>'], // force all headers to the same level
-        [/<p(.*?)>/ig, '¶'], // get rid of <p> tags
+        [/<p(.*?)>/ig, '¶'], // mark paragraphs for splitting
+        [/<br(.*?)><br(.*?)>/ig, '¶'], // mark double line breaks for splitting
         [/<\/p>/ig, '']
       ]
     },
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -10,7 +10,8 @@ var _ = require('lodash'),
   edit = require('../services/edit'),
   model = require('text-model'),
   site = require('../services/site'),
-  refAttr = references.referenceAttribute;
+  refAttr = references.referenceAttribute,
+  paragraphSplitter = '¶¶';
 
 /**
  * whan that sellection, with his ranges soote
@@ -36,7 +37,7 @@ function selectAfter(node) {
  * @returns {array}
  */
 function splitParagraphs(str) {
-  return str.split('¶');
+  return str.split(paragraphSplitter);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


behaviors/wysiwyg.js
```diff
@@ -13,6 +13,69 @@ var _ = require('lodash'),
...
+  // that hem hath holpen, when that they were pasted
+}
+
+/**
+ * split innerHTML into paragraphs
+ * @param {string} str
+ * @returns {array}
+ */
+function splitParagraphs(str) {
+  return str.split('¶');
```
cruzanmo: A far-edge case, but what happens if editorial uses the ¶ character?

yoshokatana: If they type that character, everything's fine. If they paste in html that contains that character, it'll make a new paragraph. Hmmmmm

yoshokatana: I'm gonna set it to use `'¶¶'` instead.

yoshokatana: actually, splitting on the closing paragraph tag gives us the best support for google docs, pages, and word, and doesn't require weird characters

behaviors/wysiwyg.js
```diff
@@ -270,6 +333,47 @@ function addComponent(el, text) {
...
+
+  // create the new components
+  return Promise.all(_.map(components, function (component) {
+    var throwawayDiv = document.createElement('div'),
+      fragment = model.toElement(component);
+
+    throwawayDiv.appendChild(fragment);
+
+    return edit.createComponent(current.name, {
+      text: throwawayDiv.innerHTML
```
byronhulcher: I think you can use `fragment.outerHTML` and dont need to create a random div

behaviors/wysiwyg.js
```diff
@@ -13,6 +13,71 @@ var _ = require('lodash'),
...
+  var range = document.createRange(),
+    selection = window.getSelection();
+
+  range.setStartAfter(node); // set the caret after this node
+  range.collapse(true);
+  selection.removeAllRanges();
+  selection.addRange(range);
+  // Of the DOM, to inputte's end they wende
+  // the hooly blisful caret for to seke
+  // that hem hath holpen, when that they were pasted
```
byronhulcher: Is this new es6 syntax? I'm having trouble reading this :eyes:

yoshokatana: ES1475 actually

behaviors/wysiwyg.js
yoshokatana: > If the element has no parent node, that is if it is the root node of the document, setting its outerHTML property will throw a DOMException with the error code NO_MODIFICATION_ALLOWED_ERR.

Yeah, it's not great, but apparently the temp div is the current best practice :disappointed:

byronhulcher:
:+1: to me!

cruzanmo:
+1

PR #424, opened 2016-04-15
some rules around behaviors

yoshokatana:
* `get-input` was being used to both get the actual `input` element (for logic purposes) _and_ to get the `rv-field` element (where the data is stored) for positioning purposes. This splits that into two seperate functions and renames the service to `field-helpers`
* pulled `label` styles into the actual label behavior


byronhulcher:
:+1:

PR #425, opened 2016-04-18
updated field helpers, focus

yoshokatana:
* removed `.simple-list` from `isInput` (I should have removed it when I created `getField`)
* updated field-helpers tests
* updated `focus` service to properly focus on the first input when opening a form

### DIFF
--- a/services/field-helpers.js
+++ b/services/field-helpers.js
@@ -11,7 +11,7 @@ function getInput(el) {
 }
 
 function isInput(el) {
-  return !!el && (el.nodeName === 'INPUT' || el.nodeName === 'TEXTAREA' || el.classList.contains('wysiwyg-input') || el.classList.contains('simple-list'));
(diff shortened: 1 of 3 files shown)


davidandrewfrey:
👍

services/field-helpers.js
```diff
@@ -11,7 +11,7 @@ function getInput(el) {
 }
 
 function isInput(el) {
-  return !!el && (el.nodeName === 'INPUT' || el.nodeName === 'TEXTAREA' || el.classList.contains('wysiwyg-input') || el.classList.contains('simple-list'));
+  return !!el && (el.nodeName === 'INPUT' || el.nodeName === 'TEXTAREA' || el.classList.contains('wysiwyg-input'));
```
ughitsaaron: This feels like a lot of logic to cram into one expression.

PR #432, opened 2016-04-19
set caret when clicking into forms

yoshokatana:
[trello ticket](https://trello.com/c/mYMUZSBy/309-set-cursor-when-clicking-into-inline-forms-textedit)

### DIFF
--- a/decorators/focus.js
+++ b/decorators/focus.js
@@ -2,10 +2,52 @@ var _ = require('lodash'),
   references = require('../services/references'),
   forms = require('../services/forms'),
   select = require('../services/select'),
+  sr = require('selection-range'),
   dom = require('@nymag/dom'),
   helpers = require('../services/field-helpers'),
   componentList = require('./component-list'),
   currentFocus;
+
+/**
+ * get the text offset from a click event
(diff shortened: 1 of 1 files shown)


decorators/focus.js
```diff
@@ -2,11 +2,53 @@ var _ = require('lodash'),
...
+      textNode = range.startContainer;
+      offset = range.startOffset;
+    }
+
+    parent = dom.closest(textNode, '[data-uri]'); // get component el
+    parentText = parent.textContent.replace(/^(\n|.)*?Delete\s*/, ''); // remove component selector text
+    parentOffset = parentText.indexOf(textNode.textContent) + offset;
+
+    return parentOffset;
+  } catch (e) {
```
ughitsaaron: Don't need to pass in `e` if you're not using it.

yoshokatana: according to the linter it's actually required for try/catch statements, even if you don't actually use it 😅

byronhulcher:
👍 this looks good to me.This could be a useful bit of code to pull out and share with other people through a module or similar.

PR #433, opened 2016-04-20
New page type selector

davidandrewfrey:
Creates a pane to give writers options for creating pages of various types:
https://trello.com/c/3OIIu04o/6-m-ability-to-select-a-layout-when-creating-a-new-page

Updated the way the panes are generated. Pane templates are located at the bottom of the template.nunjucks and then cloned and populated as needed.


services/pane.js
```diff
@@ -148,6 +151,28 @@ function openPublish() {
...
+
+  return state.get().then(function () {
+    // append actions to the doc fragment
+    innerEl.appendChild(pageActionsSubTemplate);
+
+    // create the root pane element
+    el = open(header, innerEl);
+    // init controller for publish pane
+    ds.controller('pane-new-page', newPagePaneController);
+    ds.get('pane-new-page', el.querySelector('.new-page-actions'));
```
cruzanmo: Could use `pageActionsSubTemplate` rather than querySelector here?

services/pane.js
```diff
@@ -91,36 +90,40 @@ function createPublishMessages(res) {
  * @returns {Element}
  */
 function createPublishActions(res) {
-  var today = moment().format('YYYY-MM-DD'),
+  var actions = dom.find('.publish-actions'),
```
cruzanmo: `.publish-actions` could be a const string at the top of the service

cruzanmo: hmm, actually, I guess all of the template are within `kiln-pane-template-elements` so that should be shared?

template.nunjucks
```diff
@@ -47,16 +47,62 @@
...
+          <button class="close">{% include 'public/media/components/clay-kiln/close.svg' ignore missing %}</button>
         </header>
         <div class="pane-inner"></div>
       </div>
     </div>
   </template>
+
+  <div class="kiln-pane-template-elements">
+
+  {# sub-templates: messages #}
```
cruzanmo: Confused me slightly that this comment is not indented, but do see your intention.

controllers/pane-new-page.js
```diff
@@ -0,0 +1,30 @@
...
+      '.create-article-page click': 'onNewArticle'
+    },
+
+    onNewArticle: function (e) {
+      e.preventDefault();
+
+      return edit.createPage()
+        .then(function (url) {
+          location.href = url;
+        })
```
cruzanmo: L19-L22 is the same as https://github.com/nymag/clay-kiln/blob/new-page-type-selector/controllers/kiln-toolbar.js#L32-L34

services/pane.js
davidandrewfrey: Definitely. Making the update.

services/pane.test.js
```diff
@@ -156,54 +157,107 @@ describe(dirname, function () {
...
-        template.content.appendChild(headEl);
-        template.content.appendChild(paneInnerEl);
-        document.body.appendChild(template);
-
-        // add toolbar
-        toolbar.classList.add('kiln-toolbar');
-        document.body.appendChild(toolbar);
+        env = new nunjucks.Environment();
+        // satisfy request for nunjucks filter in template
+        env.addFilter('includeFile', noFilter); // TODO – include nunjucks filters
```
cruzanmo: What does this do?

template.nunjucks
davidandrewfrey: Ha – yeah. Updated.

services/pane.js
davidandrewfrey: #1 – Could be a const for sure.
#2 – They all live there inside of `kiln-pane-template-elements` and could be shared, but each is directly selected.

services/pane.test.js
davidandrewfrey: Satisfies the test's hunger for nunjucks filters. We currently only need the dom elements so they aren't necessary, but it's something we need to figure out.

services/pane.test.js
```diff
@@ -156,54 +157,107 @@ describe(dirname, function () {
...
+        getState.returns(Promise.resolve({}));
+        lib.close();
+        document.body.innerHTML += result;
+        el = document.querySelector('.kiln-toolbar-button.new');
+        spy = sandbox.spy(el, 'click');
+        el.click();
+
+        function clickToOpenPane() {
+          expect(el).to.exist;
+          sinon.assert.called(spy);
```
cruzanmo: I think this will always be true because `el.click();` is run above.

services/pane.js
davidandrewfrey: #1 (update) looking at support for the client side and `const` is still a bit problematic. Safari likes in non-strict mode, chrome likes w/ strict. Moving forward with `var` for now.

services/pane.test.js
```diff
@@ -156,54 +157,107 @@ describe(dirname, function () {
...
+        clickToOpenPane();
+      });
+
+      it('opens a create page pane and clicks new article page button', sinon.test(function () {
+        lib.close();
+        document.body.innerHTML += result;
+        el = document.querySelector('.new-page-actions .create-article-page');
+        stub = sandbox.stub(el, 'click');
+        el.click();
+        sinon.assert.called(stub);
```
cruzanmo: Same as above

services/pane.test.js
```diff
@@ -156,54 +157,107 @@ describe(dirname, function () {
...
+      }));
+
+      it('creates a new article page', sinon.test(function () {
+        createPage = sandbox.stub(edit, 'createPage');
+        lib.close();
+        expect(createPage.returns(Promise.resolve({}))).to.exist;
```
cruzanmo: This is testing sinon's stub returns method... edit.createPage is not called (unless it is called by lib.close()?)

services/pane.test.js
davidandrewfrey: It's dependent on el existing, but that's nonsense! Thanks for the catch – updating.

services/pane.js
yoshokatana: you can (and should) use `let` and `const` in kiln

controllers/kiln-toolbar.js
```diff
@@ -15,11 +16,23 @@ var EditorToolbar,
  * @returns {Promise}
  */
 function createPage() {
-  // todo: allow users to choose their layout / components
```
yoshokatana: this todo should be added back, since we still don't have that functionality

controllers/pane-new-page.js
yoshokatana: that's fine

services/pane.js
```diff
@@ -91,36 +90,40 @@ function createPublishMessages(res) {
...
+  if (actions) {
+    scheduleDate = dom.find(actions, '#schedule-date');
+    scheduleTime = dom.find(actions, '#schedule-time');
+
+    if (scheduleDate) {
+      dom.replaceElement(scheduleDate, dom.create(`<input id="schedule-date" class="schedule-input" type="date" min="${today}" value="${today}" placeholder="${today}"></input>`));
+    }
+
+    if (scheduleTime) {
+      dom.replaceElement(scheduleTime, dom.create(`<input id="schedule-time" class="schedule-input" type="time" value="${now}" placeholder="${now}"></input>`));
```
yoshokatana: wait, aren't these already in the dom? Why are we replacing elements rather than just updating their attributes?

services/pane.js
```diff
@@ -148,6 +151,28 @@ function openPublish() {
   });
 }
 
+/**
+ * open new page type dialog pane
+ * @returns {Promise}
+ */
+function openNewPage() {
+  var header = 'Select New Page Type',
```
yoshokatana: How about just "New Page"?

services/pane.js
```diff
@@ -148,6 +151,28 @@ function openPublish() {
...
+    innerEl = document.createDocumentFragment(),
+    pageActionsSubTemplate = dom.find('.new-page-actions').cloneNode(true),
+    el;
+
+  return state.get().then(function () {
```
yoshokatana: Why are you calling `state.get()` if you're not using any of the page state?

services/pane.js
```diff
@@ -156,17 +181,31 @@ function addPreview(preview) {
...
-          <span class="description">${error.rule.description}</span>
-          <ul class="errors"></ul>
-        </div>
-      `),
+    var errorEl = dom.find('.publish-errors'),
+      errorLabel = dom.find(errorEl, '.label'),
+      errorDescription = dom.find(errorEl, '.description'),
       list = dom.find(errorEl, 'ul');
 
+    // reset template label default if available
```
yoshokatana: I'm not sure what this means. "template label default"? Shouldn't the template _never_ have a label or description?

services/pane.js
davidandrewfrey: Good to know! Will add.

styleguide/pane.scss
```diff
@@ -7,6 +7,11 @@
 $pane-margin: 30vh;
 $easeOutExpo: cubic-bezier(.190, 1.000, .220, 1.000);
 
+.hide,
+.kiln-pane-template-elements {
+  display: none !important; // Always hide your skeletons and other hidden things
```
yoshokatana: remove this !important tho

template.nunjucks
```diff
@@ -17,23 +17,23 @@
       data-site-assetpath="{{ site.assetPath }}"
       data-components="{{ _components }}">
       <div class="kiln-toolbar-inner">
-        <button class="user-icon">{% include 'public/media/components/clay-kiln/user-icon.svg' %}</button>
+        <button class="user-icon">{% include 'public/media/components/clay-kiln/user-icon.svg' ignore missing %}</button>
```
yoshokatana: why `ignore missing`? If these icons are missing, we've broken something horribly.

services/pane.js
davidandrewfrey: It's less code & less complicated to go this route. Fewer selects and inserts.

services/pane.js
davidandrewfrey: Yeah! Too wordy. And "type" is robotic. Since there are options, how about "Select New Page" since we are prompting the user to make a decision?

services/pane.js
davidandrewfrey: Yeah. No. Don't need it. Thanks for the catch.

services/pane.js
davidandrewfrey: Messed around with this some more. Prefer "New Page" too, as it's fast. But those words as a design element feel a little off.

controllers/kiln-toolbar.js
davidandrewfrey: Added back

services/pane.test.js
davidandrewfrey: Revised per our conversation.

services/pane.test.js
davidandrewfrey: Updated.

controllers/pane-new-page.js
cruzanmo: ok, fine. if you are looking for a trickier way:
```
edit.createPage()
  .then(_.partial(_.set, window.location, 'href'))
```

davidandrewfrey: Like this style @cruzanmo, but what we have is more inline with the codebase.

services/pane.js
davidandrewfrey: For readability, the template skeleton in the nunjucks file has a value. Also, if there is a problem with populating the error, but we still have an error condition where this dialog pane would be displayed, the error would return something to the user instead of being blank.

styleguide/pane.scss
davidandrewfrey: Yeah – I generally steer clear of `!important`. These are utility classes that need to be dependable.

template.nunjucks
davidandrewfrey: Was using these to pacify the tests as the media is currently inaccessible from the tests. Agreed this isn't great. Ideas on how better to make this functional for testing?

services/pane.js
cruzanmo: The label value should either be handled in the template or in the js. If the template needs it for readability, could use an HTML or nunjucks comment.

services/pane.test.js
```diff
@@ -156,54 +157,108 @@ describe(dirname, function () {
...
+        env.addFilter('includeFile', noFilter); // TODO – include nunjucks filters
+        template = env.getPreprocessedTemplate('template.nunjucks');
       });
 
       beforeEach(function () {
+        result = template.render(mock);
         sandbox = sinon.sandbox.create();
         sandbox.stub(ds);
-        sandbox.stub(state, 'get');
+        getState = sandbox.stub(state, 'get');
```
cruzanmo: now that `state.get` was removed from the code, I think it can be removed from the related tests as well.

styleguide/pane.scss
cruzanmo: hmm, I could go either way on this. When is it necessary to use `!important`? I would feel better not using until it reveals itself as being necessary.

services/pane.test.js
davidandrewfrey: Yeah, totally. Thanks for catching.

services/pane.js
davidandrewfrey: True @cruzanmo, _but_ this also provides a backup incase of problem. AND when reading the HTML itself, it becomes clear what those elements combine to be at-a-glance.

styleguide/pane.scss
davidandrewfrey: We should have "Utility Classes". We also need to get organized on how we are using styles. Like – how do we use multiclass selector styles to describe states? How would we name such things, etc. I know the CSS stuff is maybe not top of list but we need to get all of this together logically, especially if we introduce user-addressable-component-level-styles.

cruzanmo:
+1. Looks like the branch needs to be rebased.

cruzanmo:
ha, beware of the only

services/pane.js
yoshokatana: If we're creating the elements in js anyway, then why are we also creating them in the template? Those elements will _always_ be replaced

styleguide/pane.scss
yoshokatana: `.hide` is a SUPER generic class name that will probably be used by other things, causing bad side effects.

PR #434, opened 2016-04-20
added codemirror behavior with scss

yoshokatana:
this allows us to use a nicely-formatted text area to add styles. In the future, we can use the same behavior to do:

* yaml for custom searches
* linting of styles
* lots of other neat things

screen shot 2016-04-20 at 5 32 50 pm


ughitsaaron:
👍

PR #438, opened 2016-04-29
Revert "New page type selector"

yoshokatana:
Reverts nymag/clay-kiln#433


yoshokatana:
👎  david's gonna make some more updates in a separate pr. no need to revert

PR #439, opened 2016-04-29
Code review updates re:Nelson.

davidandrewfrey:
Keeping the skeleton/clone pattern for Pane elements. This pattern is a test drive of sorts. We may change this back to the functional pattern.

### DIFF
--- a/services/pane.js
+++ b/services/pane.js
@@ -6,7 +6,8 @@ var _ = require('lodash'),
   site = require('./site'),
   paneController = require('../controllers/pane'),
   newPagePaneController = require('../controllers/pane-new-page'),
-  publishPaneController = require('../controllers/publish-pane');
+  publishPaneController = require('../controllers/publish-pane'),
+  kilnHideClass = 'kiln-hide';
 
 /**
  * create pane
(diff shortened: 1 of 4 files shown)


services/pane.js
```diff
@@ -85,7 +85,7 @@ function createPublishMessages(res) {
     scheduleMessage = dom.find(messages, '.publish-schedule-message');
     if (scheduleMessage) {
       scheduleMessage.innerHTML = `Scheduled to publish ${state.formatTime(res.scheduledAt)}.`;
-      scheduleMessage.classList.remove('hide');
+      scheduleMessage.classList.remove('kiln-hide');
```
cruzanmo: make `'kiln-hide'` a var or const at the top

styleguide/_mixins.scss
```diff
@@ -9,3 +9,7 @@
     clear: both;
   }
 }
+
+@mixin kiln-hide {
```
cruzanmo: should mixin just be called "hide" as it could be used in other classes?

davidandrewfrey: It could, but since it is tied to this test of the pattern used for the Kiln Pane, it might be best to keep it less global.

yoshokatana:
+1

cruzanmo:
+1

davidandrewfrey:
Much appreciated @yoshokatana + @cruzanmo

PR #443, opened 2016-05-02
run validation when clicking toolbar button

yoshokatana:
* [trello ticket](https://trello.com/c/YDSVJbc7/314-allow-users-to-validate-pages-before-publishing)
* refactors some pane stuff
* removes nunjucks dependency from tests (it wasn't testing things we needed to test)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -8,8 +8,7 @@ files = [
   'behaviors/*.js',
   'services/**/*.js',
   'decorators/*.js',
-  'validators/*.js',
-  'template.nunjucks' // preprocess nunjucks templates
+  'validators/*.js'
 ];
 
 settings = {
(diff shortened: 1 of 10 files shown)


karma.conf.js
```diff
@@ -41,14 +40,13 @@ settings = {
   },
   reporters: ['dots', 'coverage'],
   files: files,
-  frameworks: ['mocha', 'chai', 'sinon', 'browserify', 'nunjucks'],
+  frameworks: ['mocha', 'chai', 'sinon', 'browserify'],
```
davidandrewfrey: Sort of sad to see this go, but per our conversation I agree that we should accomplish the dream of testing the full FE functionality from another direction.

services/pane.test.js
```diff
@@ -1,16 +1,75 @@
...
+  </div>`);
+}
+
+function stubErrorsTemplate() {
+  return dom.create(`<div class="publish-error">
+    <span class="label">There was a problem:</span>
+    <span class="description">Please see below for details</span>
+    <ul class="errors"></ul>
+  </div>`);
+}
```
davidandrewfrey: If we do accomplish that dream (mentioned above), maybe we can rethink this level of dom-y testing.

davidandrewfrey:
👍

PR #447, opened 2016-05-03
WARNING messaging

yoshokatana:
* [trello ticket](https://trello.com/c/YDSVJbc7/314-allow-users-to-validate-pages-before-publishing)
* adds `warning` validation
* BREAKING CHANGE TO THE VALIDATORS API (see https://github.com/nymag/sites/pull/981)

Warnings appear in the publish pane...

screen shot 2016-05-03 at 3 04 49 pm

And in the error pane!

screen shot 2016-05-03 at 3 05 36 pm


services/pane.js
```diff
@@ -215,56 +225,59 @@ function addPreview(preview) {
 
 /**
  * format and assemble error messages
- * @param {Object[]} errors
+ * @param {Object[]} errors (or warnings)
+ * @param {string} [modifier] modifier class for warnings, info, etc
  * @returns {Element}
  */
-function addErrors(errors) {
+function addErrors(errors, modifier) {
```
davidandrewfrey: Like what's going on with all of this. Very clear, except would like to have this function renamed something like `addErrorsOrWarnings()` to keep that clarity.

services/pane.js
```diff
@@ -215,56 +225,59 @@ function addPreview(preview) {
...
- * @param {string} errors[].rule.description e.g. 'Required fields cannot be blank'
- * @param {Object[]} errors[].errors
- * @param {string} errors[].errors[].label e.g. 'Article > Header'
- * @param {string} [errors[].errors[].preview] e.g. 'text in a paragraph TK more text...'
+ * @param {object} validation
+ * @param {Object[]} validation.errors
+ * @param {Object[]} validation.warnings
  */
-function openValidationErrors(errors) {
+function openValidationErrors(validation) {
```
davidandrewfrey: This function is being used for both errors and warnings, right? If it is – would be nice to have it named something more generic like `openValidationMessages()`.

services/pane.js
yoshokatana: good idea

services/pane.js
yoshokatana: it displays both errors and warnings, but will only open when there are errors (the warnings are sorta secondary)

services/pane.test.js
```diff
@@ -196,6 +198,42 @@ describe(dirname, function () {
...
+        fn().then(expectRegularPane);
+      });
+
+      it('adds warning message if warnings are passed in', function () {
+        getState.returns(Promise.resolve({}));
+        lib.close();
+
+        function expectRegularPane() {
+          expect(document.querySelector('.pane-header').innerHTML).to.equal('Schedule Publish');
+          expect(document.querySelector('.pane-inner .publish-valid')).to.equal(null);
```
davidandrewfrey: Still spinning up on tests. For Chai it looks like this could also end with `('.pane-inner .publish-valid')).to.not.exist` which looks for `null` and `undefined`. Do we care?

yoshokatana: Not so much. Also I like avoiding the chai methods that aren't really methods, since linters usually bug out when they see those

styleguide/pane.scss
```diff
@@ -96,20 +96,32 @@ $easeOutExpo: cubic-bezier(.190, 1.000, .220, 1.000);
...
+  position: relative;
+
   p {
     color: $black-50;
     margin: 0 0 10px;
   }
 }
 
+.kiln-toolbar-pane .publish-messages:before {
+  border-top: 1px solid $grey;
```
davidandrewfrey: Thoughts on having a modifier on `$grey`? Expect we may wind up with a few. Also, may be too early for this. We need to have a big think on what the next version of Kiln should look and feel like.

styleguide/pane.scss
```diff
@@ -217,6 +229,18 @@ $easeOutExpo: cubic-bezier(.190, 1.000, .220, 1.000);
   }
 }
 
+// special color for warnings
+.kiln-toolbar-pane .publish-warning .label {
+  color: $orange;
```
davidandrewfrey: Same comment as above for `$orange`. Feel like we're in super-flux-mode right now so maybe something to keep in back-of-mind.

services/pane.js
davidandrewfrey: Ah! OK.

services/pane.test.js
davidandrewfrey: Good to know. I'll follow the `null` way above.

davidandrewfrey:
It's nice that. +1

styleguide/pane.scss
yoshokatana: Yeah, I think we're ok for now

styleguide/pane.scss
yoshokatana: Haha yep

PR #458, opened 2016-05-18
enable layout editing

yoshokatana:
* [trello ticket](https://trello.com/c/gFg4meDF/324-full-layout-editing)
* this enables _layout_ editing, but not _page area_ editing
* note: to use this, you'll need to make sure your layout `schema` is correct, as well as adding `data-editable` to the sections you want to be editable (https://github.com/nymag/sites/pull/1088)


byronhulcher:
👍

PR #459, opened 2016-05-19
added layout selector colors

yoshokatana:
page colors will apply to components in the page:

screen shot 2016-05-19 at 12 07 21

layout colors will apply to components in the layout:

screen shot 2016-05-19 at 12 07 38

We should use these colors in the publish pane to reinforce the visual language.


byronhulcher:
👍

PR #473, opened 2016-05-31
add support for splitting paragraphs via divs

yoshokatana:
When pasting multiple paragraphs, google docs and editors like IA Writer sometimes use divs rather than `<p>` tags to denote paragraphs. This allows us to paste those correctly.

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -36,10 +36,13 @@ function selectAfter(node) {
  * @returns {array}
  */
 function splitParagraphs(str) {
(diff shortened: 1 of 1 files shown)


jonwinton:
👍

byronhulcher:
👍  tho it could be worth tying these together using a variable

yoshokatana:
@byronhulcher they're slightly different, since one is the opening and one is the closing, but if a third splittable tag is added I'll pull them out into a variable

davidandrewfrey:
👍

PR #474, opened 2016-05-31
grab available new pages from list

yoshokatana:
Instead of hardcoding two possible new pages, sites should specify which pages they want using a `lists` endpoint.

```
/lists/new-pages
```

When the new page pane opens, it'll grab the list and create buttons for any page the site allows.

screen shot 2016-05-31 at 1 38 42 pm


davidandrewfrey:
⭐ ⭐ ⭐ ⭐ ⭐

PR #477, opened 2016-05-31
magic-button can generate slugs now

yoshokatana:
magic-button should have a general slugification transform

### DIFF
--- a/behaviors/magic-button.js
+++ b/behaviors/magic-button.js
@@ -145,6 +145,7 @@ function setFieldData(bindings, field, data) {
 
 /**
  * get data from an API
+ * note: we export this and then use the exported method, so we can stub it in our tests
  * @param {string} endpoint
  * @returns {Promise}
  */
@@ -171,10 +172,12 @@ function getProperty(property) {
  * do magic on click ☆.。.:*・°☆.。.:*・°☆.。.:*・°☆.。.:*・°☆
  * @param {MouseEvent} e
  * @param {object} bindings
+ * @param {Element} [testEl] for testing, we pass a stubbed element in rather than the event
+ * note: when this function is called from the binding, testEl is undefined
  * @returns {Promise|undefined}
  */
-function doMagic(e, bindings) {
-  const el = e.currentTarget,
+function doMagic(e, bindings, testEl) {
+  const el = testEl || e.currentTarget,
     currentField = el.getAttribute('data-magic-currentField'),
     field = el.getAttribute('data-magic-field'),
     component = el.getAttribute('data-magic-component'),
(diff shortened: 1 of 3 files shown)


behaviors/magic-button.js
```diff
@@ -30,10 +43,46 @@ const _ = require('lodash'),
...
+      });
     }
   };
 
 /**
+ * Removes all unicode from string
+ * @param {string} str
+ * @returns {string}
+ */
+function stripUnicode(str) {
```
davidandrewfrey: Do we have a more Util-y place for this guy to live? Could see this being handy elsewhere.

behaviors/magic-button.js
```diff
@@ -30,10 +43,46 @@ const _ = require('lodash'),
...
+  return str.replace(/[^A-Za-z 0-9\.,\?!@#\$%\^&\*\(\)-_=\+;:<>\/\\\|\}\{\[\]~]*/g, '');
+}
+
+/**
+ * remove all html stuff from a string
+ * @param {string} str
+ * @returns {string}
+ */
+function toPlainText(str) {
```
davidandrewfrey: Same Util comment for this function too.

behaviors/magic-button.js
yoshokatana: not really. if we end up using it anywhere else, I'll think about moving it

behaviors/magic-button.js
yoshokatana: same answer :p

behaviors/magic-button.js
```diff
@@ -30,10 +43,46 @@ const _ = require('lodash'),
...
+  // coerce all text into a string. Undefined stuff is just an empty string
+  if (!_.isString(str)) {
```
davidandrewfrey: Our preference is to skip ternary operators for things like this?

yoshokatana: yeah, this is clearer

behaviors/magic-button.js
```diff
@@ -122,10 +172,12 @@ function getProperty(property) {
  * do magic on click ☆.。.:*・°☆.。.:*・°☆.。.:*・°☆.。.:*・°☆
  * @param {MouseEvent} e
  * @param {object} bindings
+ * @param {Element} [testEl] for testing, we pass a stubbed element in rather than the event
+ * note: when this function is called from the binding, testEl is undefined
  * @returns {Promise|undefined}
  */
-function doMagic(e, bindings) {
-  const el = e.currentTarget,
+function doMagic(e, bindings, testEl) {
```
davidandrewfrey: The name is 🌟 🌟 🌟 🌟 🌟

behaviors/magic-button.js
davidandrewfrey: Sounds good.

behaviors/magic-button.js
davidandrewfrey: Cool. I'll keep it in mind for consistency's sake. Feel like ternarys are fine as long as their either semi-complex AND only one level deep (like the second half would be below.) Or 2 levels deep if all moving parts are simple.

davidandrewfrey:
👍

PR #479, opened 2016-06-07
always open pane when hitting new page

yoshokatana:
[trello ticket](https://trello.com/c/TsNep1mF/351-toolbar-icons-should-never-change-system-state)


cruzanmo:
+1, less is more

PR #480, opened 2016-06-09
moved related services into folders

yoshokatana:
This pr is just a cleanup that moves related services into folders:

* `forms` for all the form-related stuff
* `field-helpers` for services that are used by behaviors to do stuff while forms are open (more will be added in here as we add more built-in behaviors)
* `components` for services that specifically deal with rendering and adding stuff to components as a whole


davidandrewfrey:
💯 
Nice to have these organized logically.

PR #482, opened 2016-06-15
kiln style tweaks

yoshokatana:
* removed old add-component styles
* fixed tiny bug in panes with buttons that aren't wide enough
* fixed bugs in component lists which exist inside of flexed containers

screen shot 2016-06-15 at 4 27 08 pm


services/components/select.js
```diff
@@ -75,30 +75,16 @@ function select(el) {
 }
 
 /**
- * hide ALL component lists in element
- * @param {Element} el
- */
-function hideComponentList(el) {
-  var lists = dom.findAll(el, '.component-list-bottom');
```
yoshokatana: these elements (the old add component stuff) don't exist anymore

styleguide/add-component.scss
```diff
@@ -1,95 +0,0 @@
...
-
-.add-component {
-  @include button-outlined($blue, #fff);
-
-  margin: 0 15px 15px 0;
-  max-width: 100%;
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-}
```
yoshokatana: _all_ of these styles deal with the old add-components stuff, not the new pane

byronhulcher:
👍

yoshokatana:
Actually, setting it to `display: flex` was dumb, and caused other issues. A simple `display: inherit` fixes everything everywhere for all time.

davidandrewfrey:
👍

PR #485, opened 2016-06-21
catch html trying to be pasted as escaped strings

yoshokatana:
I noticed, when diagnosing [this trello ticket](https://trello.com/c/NlJAHZ3U/42-bug-facebook-embed), that people were pasting iframes into paragraphs where they shouldn't be able to. `medium-editor` cleans pasted tags (including iframes), but will miss those tags if they're pasted in as escaped strings (e.g. `&lt;iframe&gt;`). This catches any escaped tags and removes them when pasting, but allows real tags to be pasted like normal.

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -132,7 +132,8 @@ function createEditor(field, buttons) {
       ],
       preCleanReplacements: [
         [/<h[1-9]>/ig, '<h2>'],
-        [/<\/h[1-9]>/ig, '</h2>'] // force all headers to the same level
+        [/<\/h[1-9]>/ig, '</h2>'], // force all headers to the same level
+        [/&lt;(.*?)&gt;/ig, ''] // catch any html trying to be sent in as escaped strings
       ]
     },
     anchor: {
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -131,9 +131,10 @@ function createEditor(field, buttons) {
         'iframe'
       ],
       preCleanReplacements: [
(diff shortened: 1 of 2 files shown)


davidandrewfrey:
👍

behaviors/wysiwyg.js
```diff
@@ -132,7 +132,8 @@ function createEditor(field, buttons) {
       ],
       preCleanReplacements: [
         [/<h[1-9]>/ig, '<h2>'],
-        [/<\/h[1-9]>/ig, '</h2>'] // force all headers to the same level
+        [/<\/h[1-9]>/ig, '</h2>'], // force all headers to the same level
+        [/&lt;(.*?)&gt;/ig, ''] // catch any html trying to be sent in as escaped strings
```
cruzanmo: Would it work to just convert `&lt;` to `<` and then let medium figure out if it is a html tag?

cruzanmo: also, should this come first, so that the h matching works?

yoshokatana: Hmm, let me test that and see if it works

yoshokatana: It does! Updating.

cruzanmo:
+1

behaviors/wysiwyg.js
```diff
@@ -131,6 +131,8 @@ function createEditor(field, buttons) {
         'iframe'
       ],
       preCleanReplacements: [
+        [/&lt;(.*?)&gt;/ig, '<$1>'], // catch any html trying to be sent in as escaped strings,
```
cruzanmo: I hope that `$1` replacement works here -- I assume it would...

yoshokatana: yup

PR #498, opened 2016-06-27
Added `flex` CSS

jonwinton:
Adding in `flex` so that the `component-list-inner` can grow/shrink within its container. It will now always fill its container's width.


yoshokatana:
cooooool 👍

PR #501, opened 2016-07-07
Allow placeholder to display values.

cruzanmo:
This PR allows a placeholder to display the field value or any value within a field-group, which is useful for components not visible during edit mode, e.g. third-party JavaScript.

The syntax, `${propertyName}`,  is similar to JavaScript [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

Example field-group schema:

```
_groups:
  settings:
    fields:
      - name
      - description
    _placeholder:
      text: Ad Unit ${name} ${description}
      height: 100px
      permanent: true
```

Example field schema:

```
videoId:
  _label: Video ID
  _display: inline
  _placeholder:
    text: Ooyala Video Id ${videoId}
    height: 100px
    permanent: true
  _has:
    - text
    - label
    - required
    -
      fn: description
      value: Please paste the ID of your video here from Ooyala Backlot.
```

### DIFF
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -150,18 +150,16 @@ Placeholders can display the value of a field, which is useful for components no
 videoId:
   _placeholder:
     text: Video Id ${videoId}
-    height: 100px
     permanent: true
 
-# field-group with values in placeholder text
+# group with values in placeholder text
 _groups:
-  settings:
+  group1:
     fields:
       - name
       - description
     _placeholder:
       text: Ad Unit ${name} ${description}
-      height: 100px
       permanent: true
 ```
 
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -143,7 +143,7 @@ adName:
     permanent: true
 ```
 
-Placeholders can display the value of a field, which is useful for components not visible during edit mode, e.g. third-party JavaScript. The syntax for displaying a value in the placeholder text is `${propertyName}`, which is similar to JavaScript [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). Note that `permanent: true` must be set. Example schemas:
+Placeholders can display the value of a field, which is useful for components not visible during edit mode, e.g. third-party JavaScript. The syntax for displaying a value in the placeholder text is `${fieldName}`, which is similar to JavaScript [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). Note that `permanent: true` must be set. Example schemas:
 
 ```yaml
 # single field with value in placeholder text
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -143,6 +143,28 @@ adName:
     permanent: true
 ```
 
+Placeholders can display the value of a field, which is useful for components not visible during edit mode, e.g. third-party JavaScript. The syntax for displaying a value in the placeholder text is `${propertyName}`, which is similar to JavaScript [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). Note that `permanent: true` must be set. Example schemas:
+
+```yaml
+# single field with value in placeholder text
+videoId:
+  _placeholder:
+    text: Video Id ${videoId}
+    height: 100px
+    permanent: true
+
+# field-group with values in placeholder text
+_groups:
+  settings:
+    fields:
+      - name
+      - description
+    _placeholder:
+      text: Ad Unit ${name} ${description}
+      height: 100px
+      permanent: true
+```
+
 Placeholders will display when you add `data-editable="fieldName"` in your component's template. If you're using a permanent placeholder and/or you don't want the user to click through and open a form, you can specify `data-placeholder="fieldName"` instead.
 
 When deciding how to add placeholders, keep these things in mind:
--- a/decorators/placeholder.js
+++ b/decorators/placeholder.js
@@ -39,7 +39,7 @@ function getFieldVal(path, data, fieldName) {
  * @returns {Function}
  */
 function replaceFieldName(path, data) {
-  return (match, fieldName) => getFieldVal(path, data, fieldName.trim());
+  return (match, fieldName) => getFieldVal(path, data, fieldName);
 }
 
 /**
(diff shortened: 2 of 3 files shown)


decorators/placeholder.js
```diff
@@ -6,18 +6,58 @@ var _ = require('lodash'),
   addComponentHandler = require('../services/components/add-component-handler');
 
 /**
+ *
+ * @param {Array} groupItems
+ * @param {string} propName
+ * @returns {object}
```
yoshokatana: returns object or null

decorators/placeholder.js
```diff
@@ -6,18 +6,58 @@ var _ = require('lodash'),
...
+ */
+function findPropInFieldGroup(groupItems, propName) {
+  return _.find(groupItems, groupProp => _.get(groupProp, '_schema._name') === propName) || null;
```
yoshokatana: I don't think you need to do `|| null` explicitly. doesn't `_.find` return null if nothing's found?

decorators/placeholder.js
```diff
@@ -6,18 +6,58 @@ var _ = require('lodash'),
...
+ * get the property value
+ * @param {string} path
+ * @param {object} data
+ * @param {string} propName
+ * @returns {String}
+ */
+function getPropVal(path, data, propName) {
+  var value = 'value';
+
+  return String( // always return a string
```
yoshokatana: this is semi-dangerous and kind of opaque. Suggest you find the value, then pass it through `.toString()`

decorators/placeholder.js
```diff
@@ -6,18 +6,58 @@ var _ = require('lodash'),
...
+ * @param {object} data
+ * @param {object} data._schema
+ * @returns {string}
  */
-function getPlaceholderText(path, schema) {
-  var placeholder = schema[references.placeholderProperty];
+function getPlaceholderText(path, data) {
+  var schema = data._schema,
+    placeholder = schema[references.placeholderProperty],
+    propNamePattern = new RegExp(/\$\{([^\}]+)}/ig); // allows property value in text, e.g. 'The value is ${propName}'
```
yoshokatana: use regex literal. also that regex is too permissive. we should use something more like `/\${(\s?\w+\s?)}/ig`

yoshokatana:
I like this approach, but left some comments

decorators/placeholder.js
```diff
@@ -6,18 +6,58 @@ var _ = require('lodash'),
...
- * @return {string}
+ * @param {string} path
+ * @param {object} data
+ * @param {object} data._schema
+ * @returns {string}
  */
-function getPlaceholderText(path, schema) {
-  var placeholder = schema[references.placeholderProperty];
+function getPlaceholderText(path, data) {
+  var schema = data._schema,
```
davidandrewfrey: Is the existence of `data` guaranteed?

decorators/placeholder.js
cruzanmo: null is object

decorators/placeholder.js
cruzanmo: _.find returns undefined unfortunately

decorators/placeholder.js
cruzanmo: I considered that, but something like an empty object will throw an error with .toString, e.g. `{}.toString(); // syntax error`. What is the danger in `String()`? I agree is it kind of opaque.

decorators/placeholder.js
davidandrewfrey: Think you can remove some of the regex elements. Maybe:
```/\${[^}]+}/ig```

Comment (cruzanmo) [inline decorators/placeholder.js]: Nice, `\w+` is a good idea, I will update.

File: decorators/placeholder.js (hunk 1 above)
Comment (yoshokatana) [inline decorators/placeholder.js]: technically yes, but in practical terms we care that it's different. see `_.isObject(null)`

File: decorators/placeholder.js (hunk 2 above)
Comment (yoshokatana) [inline decorators/placeholder.js]: ah, ok

File: decorators/placeholder.js (hunk 3 above)
Comment (yoshokatana) [inline decorators/placeholder.js]: Hmm, good point. How about this?

* write a comment above explaining that ("cast the value to a string. using a constructor will prevent errors being thrown" or something)
* use `new String()`

File: decorators/placeholder.js (hunk 4 above)
Comment (yoshokatana) [inline decorators/placeholder.js]: @davidandrewfrey the reason for my regex is that we only want to match alphanumerics (plus optional spaces before and after). `foo bar`, `foo-bar`, etc aren't valid property names in clay (`fooBar` is).

In fact, it might be easier if the regex was actually `/\${\s?(\w+)\s?}/ig`, because then we wouldn't have to do the `.trim()` on the property name later in the code.

File: decorators/placeholder.js (hunk 5 above)
Comment (yoshokatana) [inline decorators/placeholder.js]: yep

**Comment by yoshokatana** [comment]
Hmm. We should also document this in our [placeholder docs](https://github.com/nymag/clay-kiln/tree/master/behaviors#defining-placeholders)

**Comment by davidandrewfrey** [comment]
Looks good. Like following the template literals format that is in use elsewhere. +1 when the details are filled in.

File: behaviors/README.md [hunk 6]
```diff
@@ -143,6 +143,28 @@ adName:
...
+
+```yaml
+# single field with value in placeholder text
+videoId:
+  _placeholder:
+    text: Video Id ${videoId}
+    height: 100px
+    permanent: true
+
+# field-group with values in placeholder text
```
yoshokatana: "group", not "field-group"

behaviors/README.md
```diff
@@ -143,6 +143,28 @@ adName:
...
 ```
 
+Placeholders can display the value of a field, which is useful for components not visible during edit mode, e.g. third-party JavaScript. The syntax for displaying a value in the placeholder text is `${propertyName}`, which is similar to JavaScript [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). Note that `permanent: true` must be set. Example schemas:
+
+```yaml
+# single field with value in placeholder text
+videoId:
+  _placeholder:
+    text: Video Id ${videoId}
+    height: 100px
```
yoshokatana: you can remove the heights on these examples

behaviors/README.md
```diff
@@ -143,6 +143,28 @@ adName:
...
+    permanent: true
+
+# field-group with values in placeholder text
+_groups:
+  settings:
```
yoshokatana: settings don't get placeholders. suggest switching this to "group1"

decorators/placeholder.js
```diff
@@ -6,18 +6,58 @@ var _ = require('lodash'),
...
+}
+
+/**
+ *
+ * @param {string} path
+ * @param {object} data
+ * @returns {Function}
+ */
+function replaceFieldName(path, data) {
+  return (match, fieldName) => getFieldVal(path, data, fieldName.trim());
```
yoshokatana: now that we're disallowing whitespace in here, we can remove the `.trim()`

yoshokatana:
👍

davidandrewfrey:
👍

PR #505, opened 2016-07-07
moreMagic

### LINKED ISSUE #493: magic-button should allow for MORE magic!
Idea: Sometimes we want to do multiple things when you press a magic button, e.g. two api calls.

We should create a new argument to `magic-button` called `moreMagic`, in which you can pass an array of objects that match the normal arguments to `magic-button` (effectively making it recursable).

Here's an example where we grab lede data's image, then get the caption from mediaplay:

```yaml
fn: magic-button
component: lede
transform: getComponentInstance
url: $SITE_PREFIX
property: imgUrl
moreMagic:
  -
    transform: mediaplayUrl
    url: image-server.url/images/
    property: caption
```

yoshokatana:
fixes #493

Allow an optional array of objects with `{ transform, url, property }` to do multi-step magic! See docs for deets.

```yaml
component: feature-lede
transform: getComponentInstance
url: $SITE_PREFIX
property: imgUrl
moreMagic:
  -
    transform: mediaplayUrl (to change the image url into a string we can query mediaplay's api with)
    url: [mediaplay api url]
    property: metadata.caption
```

☆.。.:*・°☆.。.:*・°☆.。.:*・°☆.。.:*・°☆

### DIFF
--- a/behaviors/magic-button.md
+++ b/behaviors/magic-button.md
@@ -9,6 +9,7 @@ Append a magic button to an input.
 * **transform** _(optional)_ a transform to apply to the grabbed value
 * **url** _(optional)_ to get data from
 * **property** _(optional)_ to get from the returned data
+* **moreMagic** _(optional)_ to run the returned value through more transforms, api calls, etc
 
 ☆.。.:*・°☆.。.:*・°☆.。.:*・°☆.。.:*・°☆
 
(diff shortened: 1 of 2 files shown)


behaviors/magic-button.md
```diff
@@ -47,4 +51,18 @@ url: $SITE_PREFIX (this is a ~ special token ~ that evaluates to the prefix of c
...
+component: feature-lede
+transform: getComponentInstance
+url: $SITE_PREFIX
+property: imgUrl
+moreMagic:
+  -
+    transform: mediaplayUrl (to change the image url into a string we can query mediaplay's api with)
+    url: [mediaplay api url]
+    property: metadata.caption
+```
```
cruzanmo: Could there be a `magic` property to contain all magic unless there is something special about the first magic? Something like

```yaml
component: feature-lede
magic:
  -
    transform: getComponentInstance
    url: $SITE_PREFIX
    property: imgUrl
  -
    transform: mediaplayUrl (to change the image url into a string we can query mediaplay's api with)
    url: [mediaplay api url]
    property: metadata.caption
```

yoshokatana: The common case is only doing one iteration of magic in a button, so it's easier to surface those things the way I did. Plus, the properties allowed in moreMagic are a lot more restrictive.

davidandrewfrey:
⭐ ⭐ ⭐ ⭐ ⭐

PR #506, opened 2016-07-08
multi-field placeholders

### LINKED ISSUE #490: placeholders when multiple fields are empty
Right now, a placeholder for a group of fields can specify which field to check, e.g.

```yaml
_groups:
  group1:
    fields:
      - field1
      - field2
    _placeholder:
      ifEmpty: field1
```

We should allow people to specify more than one field in `ifEmpty`, since there are frequently fields that should display placeholders if both need data (for example, the url _and_ text of a link).

Specifying these options in an array doesn't allow us to define which logical operator we want, and some larger config object is burdensome, so I propose we use a simple, case-insensitive logical string:

```yaml
ifEmpty: field1 and field2
ifEmpty: field1 AND field2
# would display if both field1 and field2 are empty

ifEmpty: field1 or field2
ifEmpty: field1 OR field2
# would display if either field1 or field2 are empty
```

As a special _one more thing_, I'd like to also add support for my favorite logical operator, XOR:

```yaml
ifEmpty: field1 xor field2
ifEmpty: field1 XOR field2
# would display if only field1 or field2 are empty
```

yoshokatana:
Allow placeholders to check **TWO FIELDS** rather than just one. Great for groups, especially link `url` + `text` pairs. Use natural language to specify which properties to check.

```
field1 AND field2
field1 or field2
field1 XOR field2
```

Fixes #490

_Note:_ We're not allowing an arbitrary number of fields because that would require writing a [recursive descent parser](http://blog.oleganza.com/post/106246432/recursive-descent-parser-in-javascript).

### DIFF
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -215,7 +215,7 @@ _groups:
 
 When Kiln generates a form for a group, the `_display` (and `_placeholder`) properties of the individual fields are ignored. This means that you don't have to specify those properties in your fields if they're only used in groups (rather than referenced directly).
 
-If you add a `_placeholder` to a group, you _must_ either make it permanent (with `permanent: true`) _or_ specify a field it should check (with `ifEmpty: fieldName`). It will display the placeholder when that field is empty.
+If you add a `_placeholder` to a group, you _must_ either make it permanent (with `permanent: true`) _or_ specify fields it should check (with `ifEmpty: fieldName`). It will display the placeholder when that field is empty.
 
 ```yaml
 _groups:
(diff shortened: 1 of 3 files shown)


behaviors/README.md
```diff
@@ -231,6 +231,18 @@ _groups:
       ifEmpty: title
 ```
 
+Placeholders may check two different fields with a logical operator. This is handy for components with editable links, as you'll usually want to display a placeholder when _either_ the url or the link text are empty. Operators are case-insensitive, and you can use `AND`, `OR`, or `XOR`.
```
davidandrewfrey: Is this functionality limited to 2 fields or is it 2 or more with logical operators between each pair?

yoshokatana: limited to two fields

decorators/placeholder.js
```diff
@@ -112,17 +112,68 @@ function isFieldEmpty(data) {
...
+  tokens = ifEmpty.split(' ');
+
+  if (tokens.length === 1) {
+    // check a single field to see if it's empty
+    return isFieldAndEmpty(tokens[0], data);
+  } else if (tokens.length === 3) {
+    // check multiple fields
+    // note: for now this is limited to checking two fields
+    // e.g. foo AND bar, foo OR bar, foo XOR bar
+    return compareMultipleFields(tokens, data);
```
davidandrewfrey: Maybe rename this function `compareTwoFields` for now until it's comparing 2 or more?

yoshokatana: sounds good

davidandrewfrey:
⭐ ⭐ ⭐ ⭐ ¾

PR #510, opened 2016-07-08
site-specific components in component lists

### LINKED ISSUE #492: Limit Component List Include To Specific Sites
Some components in articles don't need to be used across all sites, how can we localize them to only specific sites? Example: rating component is used only on Grub.

yoshokatana:
fixes #492

allows you to specify that certain components in lists should be included/excluded only on certain sites

### DIFF
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -104,6 +104,21 @@ sideBarArea:
 
 Component lists don't use `_label` or `_display` (and will ignore them if you specify them), but they allow `_placeholder`. It's recommended to add placeholders to component lists, which will display when that list is empty.
 
+##### Site-specific Components
+
+You may optionally specify that components in a list should be included or excluded on a specific site. To do so, add a comma-separated list of sites in parenthesis after the name of the component in the `include` list:
+
+```yaml
+include:
+  - article (site1, site2) # allow on site1 and site2 ONLY
+  - paragraph # allow on all sites
+  - image (not:site1) # allow on all sites EXCEPT site1
+  - link (site1, site2, site3, not:site2, not:site3) # only allow on site1
+  # if (for some reason) you both include and exclude a site, it'll filter by the
+  # included sites first, then filter out the excluded. this works the same way as the
+  # include and exclude lists above
+```
+
 **Note:** In the future you will be able to [specify a minimum and maximum number of components](https://github.com/nymag/clay-kiln/issues/298) in your component lists.
 
 #### Defining Display
--- a/services/components/add-component-handler.js
+++ b/services/components/add-component-handler.js
@@ -29,19 +29,19 @@ function getParentListElement(el, path) {
 function availableOnCurrentSite(slug, logic) {
   var tokens = logic.split(',').map((str) => str.trim()), // trim any extra whitespace
     // list of site slugs to include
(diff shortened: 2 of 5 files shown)


services/components/add-component-handler.js
```diff
@@ -20,21 +21,66 @@ function getParentListElement(el, path) {
...
 /**
+ * determine if a component is available on the current site, based on logic
+ * @param {string} slug e.g. "grubstreet"
+ * @param {string} logic e.g. (di, press, thecut) or (not:di, not:press, not:thecut) or a mixture of both
+ * @returns {boolean}
+ */
+function availableOnCurrentSite(slug, logic) {
+  var tokens = logic.split(',').map((str) => str.trim()), // trim any extra whitespace
+    // list of site slugs to include
+    include = _.reject(tokens, (token) => _.includes(token, 'not:')),
```
byronhulcher: could you call these "sitesToInclude" "sitesToExclude"? Self-documenting variable names!

services/components/add-component-handler.js
```diff
@@ -20,21 +21,66 @@ function getParentListElement(el, path) {
...
-    if (exclude && exclude.length) {
-      if (!_.contains(exclude)) {
-        return item;
-      }
-    } else {
-      return item;
-    }
-  }));
+  return _.map(_.filter(possibleComponents, (item) => filterComponent(item, exclude)), function (str) {
+    return str.replace(/\s?\(.*?\)/g, ''); // remove any site logic
```
byronhulcher: This comment does not help me understand what is going on in this regex but im also dumb

cruzanmo: Could use arrow function here to be consistent.

byronhulcher:
👍 with some minor comments about comments

behaviors/README.md
```diff
@@ -104,6 +104,21 @@ sideBarArea:
...
+##### Site-specific Components
+
+You may optionally specify that components in a list should be included or excluded on a specific site. To do so, add a comma-separated list of sites in parenthesis after the name of the component in the `include` list:
+
+```yaml
+include:
+  - article (site1, site2) # allow on site1 and site2 ONLY
+  - paragraph # allow on all sites
+  - image (not:site1) # allow on all sites EXCEPT site1
+  - link (site1, site2, site3, not:site2, not:site3) # only allow on site1
```
cruzanmo: Hmm, do site-specific options belong in the schema? Seems like schema should be only tied to the component and generic to any site.

services/components/add-component-handler.js
yoshokatana: fiiiiiiiine

behaviors/README.md
yoshokatana: as a less-used +1, I think this is fine, especially since it's only for implementation-specific components (adding specific sites to a list in a component on npm would be dumb)

yoshokatana:
👍  from byron

byronhulcher:
Thanks for the improvements @yoshokatana  👍

byronhulcher:
WHOAH

byronhulcher:
WHOAH WHOAH WOOOOOOOAH

yoshokatana:
@byronhulcher @cruzanmo should we revert this?

PR #512, opened 2016-07-11
All components should be treated as editable

byronhulcher:
Components without a editable or placeholder field were not being treated as editable, which meant that kiln, among other things, did not capture and squash clicks on <a> tags within components.  Often, these components did have their <a> click squashed by nature of having a PARENT who was editable, and on page load clay kiln would squash their links too (since the add() function of our events module searches for all children matching the tag).  However, subsequently created child components would not have their link clicks squashed.

We should treat all components as editable, as they are! Even if there is not actually a thing for them to edit.

### DIFF
--- a/controllers/component-edit.js
+++ b/controllers/component-edit.js
@@ -38,18 +38,6 @@ function ComponentEdit() {
   }
 
   /**
-   * find out if component is editable or contains editable children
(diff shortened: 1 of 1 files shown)


yoshokatana:
👍

PR #521, opened 2016-07-19
sanitize blockquotes and other html entities when pasting

yoshokatana:
* strip `<blockquote>`s from pasted text, allowing users to post quotes inside quote components without them getting double quoted (and fixing an unrecoverable error when pasting them into paragraph components)
* parse ampersands, curly quotes, ellipses, and dashes pasted as html entities
* parse nonbreaking spaces into regular spaces when pasted as html entities


jonwinton:
👍

PR #522, opened 2016-07-20
fix sass imports from npm modules

yoshokatana:
As of npm v3, dependencies get flattened automatically so we can't rely on absolute paths for our codemirror (and potentially other) sass imports. This fixes the issue. Also removes responsive-filenames from our gulp script, since it's unused in kiln.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,7 +14,6 @@ var gulp = require('gulp'),
   // uglify = require('gulp-uglify'),
   concat = require('gulp-concat'),
   sass = require('gulp-sass'),
-  rfn = require('responsive-filenames'),
   autoprefix = require('gulp-autoprefixer'),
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -14,7 +14,7 @@ var gulp = require('gulp'),
   // uglify = require('gulp-uglify'),
   concat = require('gulp-concat'),
   sass = require('gulp-sass'),
-  rfn = require('responsive-filenames'),
```
byronhulcher: Don't we have to wait for all components to move over to kill rfn?

amycheng:
👍

gulpfile.js
yoshokatana: not from kiln

byronhulcher: d'oh, wrong repo

PR #530, opened 2016-07-25
site-specific-selector

### LINKED ISSUE #523: site-specific selector
```yaml
fn: site-specific-selector
sites:
  -
    url: grubstreet.com # regex
    options:
      - foo
      - bar # string array, like radio/select behavior
```

* displays as a `<select>` element, using the options for the specific site
* site is determined by a naive regex match of the url

yoshokatana:
* fixes #523
* creates `<select>` element based on site-specific options
* allows `default` options if no site matches

### DIFF
--- a/behaviors/site-specific-select.js
+++ b/behaviors/site-specific-select.js
@@ -1,0 +1,58 @@
+var _ = require('lodash'),
+  dom = require('@nymag/dom'),
+  site = require('../services/site');
+
+/**
+ * Create options for select.
+ * @param {[string]} options
+ * @returns {string}
+ */
+function createOptions(options) {
+  return _.map(options, function (option) {
+    return `<option value="${option}">${ _.startCase(option) || 'None' }</option>`;
+  }).join('\n');
+}
+
+/**
+ * create a field from some options
+ * @param {string} name
+ * @param {array} options
+ * @returns {Element}
+ */
+function createField(name, options) {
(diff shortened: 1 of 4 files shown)


behaviors/site-specific-select.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * create a field from some options
+ * @param {string} name
+ * @param {array} options
+ * @returns {Element}
+ */
+function createField(name, options) {
+  return dom.create(`
+    <label class="input-label">
+      <select class="editor-select" rv-field="${name}" rv-value="${name}.data.value">
```
ughitsaaron: you need to use backticks for the string interpolation here dogg

behaviors/site-specific-select.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param {obejct} args
+ * @param {array} args.sites
+ * @param {array} args.default
+ * @returns {{}}
+ */
+module.exports = function (result, args) {
+  var name = result.name,
+    defaultOptions = args.default,
+    currentSite = site.get('slug'),
+    matched = _.find(args.sites, (siteConfig) => siteConfig.slug === currentSite);
```
ughitsaaron: eh, you could get rid of the parens around `siteConfig`

ughitsaaron:
+1

PR #532, opened 2016-07-26
fix unpublish button not appearing

### LINKED ISSUE #524: published pages don't show unpublish button until page reload
When a page is scheduled, opening up the publish pane will display the "unschedule" button. This isn't happening with published articles and the "unpublish" button, unless you reload the page.

yoshokatana:
fixes #524

there was a race condition in `page-state` where sometimes the call to `edit.getDataOnly(pageUri)` would return the cached version, rather than doing an api call. Using the straight `db.get()` for both the published and scheduled uris fixes this.

### DIFF
--- a/services/page-state.js
+++ b/services/page-state.js
@@ -96,7 +96,7 @@ function getArticleDate(pageData) {
  * @returns {Promise}
  */
 function getPublished(publishedUri) {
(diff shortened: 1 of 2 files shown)


jonwinton:
👍

Issue #525, opened 2016-07-22
component instance names


nelsonpecora:
In certain situations, it's useful for component instances to have some kind of human readable name that can be displayed (in lists, when searching for / adding already-created instances, etc). We should figure out a standardized way to do this for all components.

yoshokatana:
Idea: similar to `_description`, a property in the component schema that specified what field (fields?) should be used for the "name", e.g.

```yaml
_name: adDescription

adDescription:
  _label: Ad Description
  _has: text
```

That would display whatever data is in the `adDescription` field as the instance's "name"

yoshokatana:
Idea: similar to placeholders, make the name a string that can interpolate fields, e.g.

```yaml
_name: Ad - ${adDescription}

adDescription:
  _label: Ad Description
  _has: text
```

This would make things like having icons (for flex logic) and stuff in the name work, as well as using data from potentially _multiple_ fields

yoshokatana:
Now that we have field interpolation, where do we actually want to use these names?

* component selectors?
* ...?

nelsonpecora:
This would be useful in the `Find Components` pane, but that's the only good use case I can think of that _isn't_ handled by the placeholder interpolation

PR #533, opened 2016-07-26
moved field interpolation to new service

yoshokatana:
* added `componentData` to options passed into decorators
* moved placeholder's field name interpolation to new service
* added tests

Now it'll be super easy to interpolate component data in any situation where we might want it, e.g. if we want _per-instance_ component names in selectors or other lists #525


jonwinton:
👍

PR #535, opened 2016-07-27
add component to property

yoshokatana:
[trello ticket](https://trello.com/c/s7SHxGMU/87-add-component-to-property)

screen shot 2016-07-27 at 12 15 23 pm

Notes:

* replace component in property
* works similarly to `_componentList` (e.g. include, exclude, auto-add if a single component is available, otherwise open the pane)
* currently you can only replace components that are already added to properties
* fixed placeholders in component lists when adding new components
* added docs for properties
* added docs for page-specific lists/props
* placeholders for properties are NOT handled yet
* component deletion in properties are NOT handled yet
* component addition to properties that don't have a component to replace are NOT handled yet


jonwinton:
👍

PR #536, opened 2016-07-27
set maxwidth for inline forms

yoshokatana:
this prevents BIG inline forms from smooshing against the edges of the viewport, since inline forms don't have any inherent padding.

before:

screen shot 2016-07-27 at 3 10 59 pm

after:

screen shot 2016-07-27 at 3 10 45 pm


jonwinton:
👍

PR #551, opened 2016-08-04
tabbed panes

nelsonpecora:
* all panes have tabs now! tabs are cool.
* you can set which active tab you want when opening panes
* you can disable tabs when opening panes
* validation/error pane is completely revamped to take advantage of this
* all panes are now the same width
* panes set their heights when opening so switching tabs won't jump them around

All panes have tabs:

screen shot 2016-08-04 at 11 13 14 am

Panes may have an optional "dynamic tab" that allows arbitrary styling and sits on the right:

screen shot 2016-08-04 at 11 12 18 am

You can specify which tab is active when opening the pane, as well as disabling certain tabs:

screen shot 2016-08-04 at 11 12 29 am


jonwinton:
👍

PR #552, opened 2016-08-04
config settings to generate html cover coverage reports locally

amycheng:
some karma config changes so that HTML code coverage reports are generated when we run `npm run test-local`.

HTML reports would provide more immediate feedback than using coveralls.io

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -22,6 +22,8 @@ settings = {
         presets: ['es2015']
       }),
       istanbul({
+        // this config fixes the following:
+        // https://github.com/karma-runner/karma-coverage/issues/157
         instrumenterConfig: {
           embedSource: true
         },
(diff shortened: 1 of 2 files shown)


karma.conf.js
```diff
@@ -22,6 +22,9 @@ settings = {
         presets: ['es2015']
       }),
       istanbul({
+        instrumenterConfig: {
```
amycheng: fixes this [bug](https://github.com/karma-runner/karma-coverage/issues/157)

byronhulcher:
:+1:

karma.conf.js
byronhulcher: Would you mind adding a comment linking to this ?

nelsonpecora:
Showing the coverage increases the coverage. There's something deeply philosophical about that. 👍

PR #553, opened 2016-08-04
kiln plugins

### LINKED ISSUE #540: proposal: kiln plugins
Should work similarly to how behaviors, decorators, and validators work. Allows people's Clay installs to change the _editing experience itself_. Useful for things like spaces and word counts.

We could set this up similarly to how behaviors, decorators, and validators work. Basically an array of objects, with each containing some functions/hooks into kiln. Some ideas:

`init` - function that runs on page load, after kiln has run its controllers. can be used to add things to selectors, toolbar, etc.

`onSave` - function that gets called after any component saves

Alternatively, we could set this up with some kind of eventing system plugins could hook into, but there might be situations where we want to pause Kiln execution until a plugin has finished running.

nelsonpecora:
* fixes #540
* adds `window.kiln.plugins` object (and service)
* adds `window.kiln` events (see readme)
* exposes certain services in `window.kiln.services`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,37 @@ Kiln has a simple client-side api that is used to extend its functionality. It c
 * [Behaviors](https://github.com/nymag/clay-kiln/tree/master/behaviors#behaviors) — run on each field when opening forms
 * [Decorators](https://github.com/nymag/clay-kiln/tree/master/decorators#decorators) — run on each field (elements with `data-editable`) when the page is loaded
 * [Validators](https://github.com/nymag/clay-kiln/tree/master/validators#validators) — run when opening the `PUBLISH` pane and returns errors and warnings
+* [Plugins](https://github.com/nymag/clay-kiln/tree/master/README.md#plugins) — run before kiln code is initialized, and can hook into events and call kiln services.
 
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -45,8 +45,37 @@ Kiln has a simple client-side api that is used to extend its functionality. It c
...
+**Events:**
+
+* `save` - after a component is saved
+* `schedule`
+* `unschedule`
+* `publish`
+* `unpublish`
+* `select` - after a component is selected
+* `unselect`
+* `add-selector` - after a selector is added to a component
```
jonwinton: Might be a good idea to describe the event data associated with the event?

services/components/select.js
```diff
@@ -431,14 +433,17 @@ function handler(el, options) {
 
     // add selector to the component el, BEHIND the component's children
     dom.prependChild(el, selector);
+    window.kiln.trigger('add-selector', el, options);
```
jonwinton: We should pass along `parent` as well. Spaces needs the context of the parent and what is in that component list to function properly.

README.md
jonwinton: We'll do this later when the system has been tested.

jonwinton:
👍

PR #555, opened 2016-08-05
make new page pane use filterable list

nelsonpecora:
this changes the new page pane to use the filterable list, rather than a list of buttons. it's the first step towards making the new page cloning more dynamic.

screen shot 2016-08-05 at 11 35 23 am

### DIFF
--- a/controllers/new-page-pane.js
+++ b/controllers/new-page-pane.js
@@ -1,31 +1,192 @@
-const edit = require('../services/edit'),
-  progress = require('../services/progress');
+var edit = require('../services/edit'),
+  dom = require('@nymag/dom'),
+  keyCode = require('keycode'),
+  _ = require('lodash'),
+  progress = require('../services/progress'),
+  pane = require('../services/pane');
+
+/**
+ * filter items in the list, based on the label or name
+ * @param {string} val
+ * @param {NodeList} items
+ */
+function filter(val, items) {
+  _.each(items, function (item) {
+    var label = item.textContent.toLowerCase(),
+      name = item.getAttribute('data-item-name').toLowerCase();
+
+    val = val.toLowerCase();
+
+    if (_.includes(label, val) || _.includes(name, val)) {
+      item.classList.remove('filtered');
+    } else {
+      item.classList.add('filtered');
+    }
+  });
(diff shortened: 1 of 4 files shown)


controllers/new-page-pane.js
```diff
@@ -1,31 +1,192 @@
...
+function getAvailable(items) {
+  return _.filter(items, (item) => !item.classList.contains('filtered'));
+}
+
+/**
+ * focus first available item in list
+ * @param {NodeList} list
+ */
+function focusFirst(list) {
+  list[0].focus();
```
byronhulcher: I think this'll throw an error if list[0] doesn't exist you should check for _isArray(list) maybe

controllers/new-page-pane.js
```diff
@@ -1,31 +1,192 @@
-const edit = require('../services/edit'),
-  progress = require('../services/progress');
+var edit = require('../services/edit'),
```
byronhulcher: y u no const

controllers/new-page-pane.js
nelsonpecora: if list[0] doesn't exist a bunch of stuff will be broken. I'm gonna add a test on instantiation when I make this generic

controllers/new-page-pane.js
nelsonpecora: I didn't notice that when changing it. I'll keep it in mind when writing the generic version

controllers/new-page-pane.js
```diff
@@ -1,31 +1,192 @@
...
+      } else if (key === 'esc') {
+        pane.close();
+      } else {
+        filter(input.value, this.items);
+      }
+    },
+
+    onItemKeydown: function (e) {
+      // simulate active states when pressing enter
+      if (keyCode(e) === 'enter') {
```
byronhulcher: I think this would be better to compare to the value from keyCode https://www.npmjs.com/package/keycode#maps

controllers/new-page-pane.js
byronhulcher: cool!

byronhulcher:
👍

PR #557, opened 2016-08-05
filterable list

nelsonpecora:
* basic version of the new filterable list service
* consolidated controller for filterable lists
* consolidated [api for creating them](https://gist.github.com/nelsonpecora/1709d900fdf8c7c6af998d4667640e21)
* this pull request only includes the `click` option, because it's used by add-component and new page
* in the near future I'll add the other (optional) options


controllers/filterable-list.js
```diff
@@ -1,8 +1,6 @@
-var edit = require('../services/edit'),
-  dom = require('@nymag/dom'),
+const dom = require('@nymag/dom'),
```
ughitsaaron: using const for importing dependencies makes sense but it's inconsistent across the rest of the project. no big, but worth considering when what we're looking for during code review, etc. if we're going to start going with `const`s then we should replicate/enforce that elsewhere in kiln code.

ughitsaaron:
looks good

controllers/filterable-list.js
nelsonpecora: Yep, we're slowly moving towards using const for these, [per Byron's comment here](https://github.com/nymag/clay-kiln/pull/555#discussion_r73737926)

PR #559, opened 2016-08-08
Additional services

jonwinton:
Exposing some additional kiln services


services/components/add-component-handler.js
```diff
@@ -133,6 +133,8 @@ function addHandler(button, options, prevRef) {
 
 module.exports = addHandler;
 module.exports.getParentEditableElement = getParentEditableElement;
+_.set(window, 'kiln.services.getAddableComponents', getAddableComponents); // export for plugins
```
nelsonpecora: why not export the `exports`?

jonwinton: Hmm I haven't needed `getParentEditableElement` so I didn't want to expose something that might never be used. But if we want to expose it I'm down.

services/edit/index.js
```diff
@@ -650,6 +658,29 @@ function getComponentRef(uri) {
   }
 }
 
+/**
+ * get the components HTML and pass along any query paramaters
```
nelsonpecora: component's, parameters

services/edit/index.js
```diff
@@ -650,6 +658,29 @@ function getComponentRef(uri) {
...
+ * that may need to be exposed to the component in it's server.js
```
nelsonpecora: its

services/edit/index.js
```diff
@@ -650,6 +658,29 @@ function getComponentRef(uri) {
...
+ * file. This is handy for passing the current page URL to the
+ * component for re-rendering when in edit mode.
+ *
+ * TODO: Currently this is ONLY used for the Space component because
+ * querying for artile tags is limited. There is a pending update
```
nelsonpecora: article

services/edit/db.js
```diff
@@ -247,6 +247,17 @@ function getHTML(uri) {
 
 /**
  * @param {string} uri
+ * @param {string} queries
+ * @return {Promise}
+ */
+function getHTMLWithQuery(uri, queries) {
+  var queryUrl = uri + extHtml + editMode + queries;
```
nelsonpecora: queries should probably be an object, `_.assign`'d to `{ edit: true}`, then stringified into `?edit=true&query1=whatever&query2=whatever` so consumers don't need to care about the ampersands/question marks

nelsonpecora:
👍  when tests pass

PR #560, opened 2016-08-08
add settings and remove options to filterable lists

nelsonpecora:
* settings
* remove

Note: remove has a confirmation, similarly to the remove button in component selectors.

Note: Also there's no keyboard shortcut for remove, partially because we don't have one for component selectors but partially because I can't distinguish between "pressing enter to confirm" and "pressing enter on a list item"


jonwinton:
👍

PR #562, opened 2016-08-08
reorder and add options

nelsonpecora:
* reorder - adds grip icons to each item, and enables dragula
* add - puts an add button (w/ title) at the bottom of the list
* addTitle - allows you to pass custom text to the title


jonwinton:
👍

PR #563, opened 2016-08-08
dynamic publishing button

nelsonpecora:
* new icons for new page, preview, and publish
* publish button has icons and text for different states
* full-height buttons and thinner toolbar
* new "save" color, brown (used for draft state and saving indicators/actions)

screen shot 2016-08-08 at 3 10 35 pm


services/page-state.js
```diff
@@ -139,16 +139,16 @@ function getPageState() {
...
-function toggleScheduled(isScheduled) {
-  var el = document.querySelector('.kiln-toolbar-inner .publish'),
-    scheduledClass = 'scheduled';
+function toggleButton(state, enabled) {
+  var el = dom.find('.kiln-toolbar-inner .publish');
 
-  if (isScheduled) {
-    el.classList.add(scheduledClass);
+  if (enabled) {
+    el.classList.add(state);
```
byronhulcher: you could replace these lines with `el.classList.togger(state, enabled)` https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

nelsonpecora:
👍  from Byron

PR #564, opened 2016-08-08
show Loading button when loading page state

nelsonpecora:
instead of showing "Draft", show "Loading..." then transition to the proper page state.

screen shot 2016-08-08 at 4 18 42 pm


byronhulcher:
👍

PR #566, opened 2016-08-08
display published state for non-article pages

nelsonpecora:
When getting the page state, we were failing the published check if a page didn't have an article. Instead, simply fall back to saying "Page is currently published." Also, this makes the language consistent between the state message and the publish pane's info message.

screen shot 2016-08-08 at 6 05 11 pm

screen shot 2016-08-08 at 6 05 50 pm


byronhulcher:
💯

PR #567, opened 2016-08-09
fixed syntax error with prevRef

nelsonpecora:
When I moved add component to use the filterable list, I thought `options.prevRef` was the property for the component we wanted to add things after. It's actually just `options.ref`.


byronhulcher:
👍

PR #568, opened 2016-08-09
fix codemirror styling issue

nelsonpecora:
fixes important codemirror styling issue which was preventing us from using it at all


gloddy:
👍

PR #569, opened 2016-08-09
use native fetch + polyfill

nelsonpecora:
This updates kiln to use native `fetch` (falling back to a polyfill on safari), rather than old-school `XMLHttpRequest`. It simplifies our codebase quite a bit, and should give us a nice little speed increase as well.


jonwinton:
👍

PR #573, opened 2016-08-10
adds startCase function that doesn't strip dash separator

ughitsaaron:
https://trello.com/c/mKd3yjBx/39-content-channels

This will maintain dash separators in content channel names.


nelsonpecora:
👍  when tests pass

PR #574, opened 2016-08-10
use default text size in selects

ughitsaaron:
https://trello.com/c/mKd3yjBx/39-content-channels

Let's not style select font-size.


nelsonpecora:
👍

PR #580, opened 2016-08-12
custom url tab

nelsonpecora:
* [trello ticket](https://trello.com/c/OnMDeM3Q/18-custom-url)
* allows typing in custom urls
* will display if a page already has a custom url
* displays the path from the beginning slash
* allows paths with/without beginning slash, or full urls
* saved to page data
* uses same form ux as adding new pages to the pages list (discrete save action, page-colored success message, etc)

screen shot 2016-08-12 at 1 51 43 pm


gloddy:
👍

nelsonpecora:
NOTE: Custom urls will not be _published_ until we make a change to sites's publishing rules

PR #584, opened 2016-08-16
handle docfragments in expectHTML

nelsonpecora:
* log error messages when `edit.save` hits an error
* handle document fragments in `db.expectHTML`
* thanks to @vicfriedman for finding this bug!

### DIFF
--- a/services/edit/db.js
+++ b/services/edit/db.js
@@ -178,10 +178,10 @@ function expectHTMLResult(uri) {
   return function (res) {
     return res.text().then(dom.create) // string -> elements
     .then(function (html) {
-      if (html.nodeType === 1) {
+      if (html.nodeType === html.ELEMENT_NODE) {
         // it's an element, add the uri
         html.setAttribute(references.referenceAttribute, uri);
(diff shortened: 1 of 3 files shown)


services/edit/db.js
```diff
@@ -176,13 +176,15 @@ function expectJSONResult(res) {
...
   return function (res) {
-    return res.text().then(function (body) {
-      // string -> elements
-      return dom.create(body);
-    })
+    return res.text().then(dom.create) // string -> elements
     .then(function (html) {
-      // add uri
-      html.setAttribute(references.referenceAttribute, uri);
+      if (html.nodeType === 1) {
```
byronhulcher: Could you use `Node type constants` instead of integers? https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#Node_type_constants

nelsonpecora: good idea

byronhulcher:
Almost wanna give you the 👍 but should this have a test?

nelsonpecora:
Yep, I'm adding one now

PR #587, opened 2016-08-17
Display cursor as pointer for relevant checkbox and radio elements...

davidandrewfrey:
...plus their labels.

No Trello card, just a quick update.


nelsonpecora:
👍

PR #589, opened 2016-08-18
put components button on the left

nelsonpecora:
this is a temporary measure so that we don't interrupt the muscle memory of users who want to create new pages. it'll be updated when we redesign the toolbar.

screen shot 2016-08-18 at 10 50 16 am


cruzanmo:
+1

PR #591, opened 2016-08-18
new preview + share pane

nelsonpecora:
* preview tab with responsive site previews
* all site previews load in new windows with defined sizes
* built in a way that allows for new sizes to be added easily in the future
* old "preview" functionality now in Shareable Link tab
* updated info text
* added copy to clipboard button with success/error messages

screen shot 2016-08-18 at 2 52 47 pm

Preview Actions

screen shot 2016-08-18 at 2 52 21 pm

Share Actions

screen shot 2016-08-18 at 2 52 34 pm


controllers/preview-pane.js
```diff
@@ -0,0 +1,56 @@
+const dom = require('@nymag/dom'),
+  _ = require('lodash'),
+  previewSizes = {
+    small: { w: 375, h: 660 },
+    medium: { w: 768, h: 1024 },
```
byronhulcher: I think our "medium" screen size in clay is 600? That's what we base our css off of at least

gloddy:
👍

controllers/preview-pane.js
nelsonpecora: our breakpoint is 600 (for the most part), but I think the most common "tablet-ish" screen width is 768

byronhulcher: That's what everyone tells me but it seems like we use them both arbitrarily :\ Another thing on the list of stuff I'd like to standardize across design + dev.

PR #592, opened 2016-08-18
larger, more angular selectors

nelsonpecora:
* [trello ticket](https://trello.com/c/T4fE50sk/21-component-selector-mvp)
* selectors are now 48px height
* updated with material design icons
* removed border radii and inner borders
* top and bottom selectors are now the same size
* tighter padding around components, to match the mocks

screen shot 2016-08-18 at 5 18 18 pm


jonwinton:
👍

gloddy:
👍

PR #593, opened 2016-08-19
Icon checkup, round 1

gloddy:
- removed retired icons
- optimized where needed
- some normalizing

More to come in a second PR...


nelsonpecora:
👍

PR #595, opened 2016-08-19
border on top of toolbar to visually separate it from content

gloddy:
Before:
Before

After:
After


nelsonpecora:
That looks a little weird...

screen shot 2016-08-19 at 1 53 44 pm

gloddy:
Hmm. Agreed.

gloddy:
@nelsonpecora Going to kill this for now till we figure out a healthy shadow pattern.

PR #596, opened 2016-08-22
validate custom urls

nelsonpecora:
* [trello ticket](https://trello.com/c/4bXk7e7o/10-validate-custom-urls-kiln)
* validates that a url is routable in a particular site
* allows full urls
* allows unsetting custom url (with improved status message)
* relies on https://github.com/nymag/amphora/pull/300

### DIFF
--- a/controllers/publish-pane.js
+++ b/controllers/publish-pane.js
@@ -1,4 +1,4 @@
-var moment = require('moment'),
+const moment = require('moment'),
   pane = require('../services/pane'),
   edit = require('../services/edit'),
   progress = require('../services/progress'),
@@ -6,7 +6,10 @@ var moment = require('moment'),
   state = require('../services/page-state'),
   db = require('../services/edit/db'),
   site = require('../services/site'),
-  _ = require('lodash');
+  _ = require('lodash'),
+  Routable = require('routable'),
+  // grab available routes from the toolbar
+  availableRoutes = dom.find('.kiln-toolbar').getAttribute('data-routes').split(',');
 
 /**
  * schedule page and layout publishing in parallel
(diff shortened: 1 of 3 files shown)


controllers/publish-pane.js
```diff
@@ -153,6 +181,21 @@ module.exports = function () {
...
+        val = input.value,
+        routes = this.routes;
+
+      // validate that what the user typed in is routable
+      // note: if it's empty string, catch it early (removing custom urls is totally valid)
+      // note: if it's a full url, assume the user knows what they're doing and say it's valid
+      if (val === '' || val.match(/^http/i) || isValidUrl(val, routes)) {
+        input.setCustomValidity('');
+      } else {
+        input.setCustomValidity('Custom URL must match an available route!');
```
davidandrewfrey: `customValidity` refers to the message here. Is this setting something that is always a string or can it be another type, like an object?

controllers/publish-pane.js
```diff
@@ -185,13 +228,19 @@ module.exports = function () {
         })
         .then(function () {
           progress.done('page');
-          progress.open('page', 'Saved custom page url', true);
+          if (url === '' || url === undefined) {
```
davidandrewfrey: A truthy test of `url` isn't explicit enough?

controllers/publish-pane.js
nelsonpecora: it's always a string. emptystring means it's valid

controllers/publish-pane.js
nelsonpecora: hmm, might be. I was being very explicit here, but I'll test it otu

davidandrewfrey:
👍

PR #598, opened 2016-08-23
invisible list

nelsonpecora:
* add, remove, reorder head components
* open settings forms for head components
* refactors pane service into a folder of specialized services
* refactors template stubbing into a fixture
* updated filterable-list icon styles
* added more tests
* updated icon styles in filterable-lists
* _BREAKING:_ client-side pane services are in `kiln.services.panes.<openWhatever>`

TODO:

* determining page vs. layout when saving stuff could be cleaned up
* determining page vs. layout when opening progress bar could be cleaned up

screen shot 2016-08-23 at 5 53 19 pm


nelsonpecora:
it's still not accurately giving the coverage, but that's a larger issue than this PR

gloddy:
Looks 👍 to me. Should get a second one from Jon though given the size.

jonwinton:
👍

PR #601, opened 2016-08-25
component selector colors

nelsonpecora:
[trello](https://trello.com/c/lgkRAUbR/28-page-layout-designation-on-selector)

screen shot 2016-08-25 at 12 40 54 pm

### DIFF
--- a/styleguide/component-select.scss
+++ b/styleguide/component-select.scss
@@ -107,22 +107,11 @@ $layout-parent-bg-color: $purple-25;
 .selected-label {
   @include label();
 
+  color: $selector-icon;
   font-size: 14px;
   line-height: $action-icon-small; // same vertical space as the icons
   margin: 0;
   white-space: nowrap;
-}
-
-// regular button styles
-.selected-label,
-.selected-info-parent,
-.selected-action-settings,
-.selected-action-delete,
-.selected-add,
-.selected-replace {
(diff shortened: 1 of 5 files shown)


styleguide/component-select.scss
```diff
@@ -101,49 +92,28 @@ $layout-parent-bg-color: $purple-25;
   justify-content: flex-start;
 }
 
+.selected-info {
+  border-left-width: 10px;
+  border-left-style: solid;
+  border-left-color: $layout-color;
```
davidandrewfrey: Minor stuff:
– alphabetical order
– could just set with `border-left` shorthand as well

nelsonpecora: we're explicitly overriding the color later, but good catch about the alpha

styleguide/component-select.scss
```diff
@@ -101,49 +92,29 @@ $layout-parent-bg-color: $purple-25;
   justify-content: flex-start;
 }
 
+.selected-info {
+  // we're explicitly overriding the color below
```
davidandrewfrey: Yeah! It'll still do that with the shorthand. Just gave it a whirl to confirm:
http://codepen.io/davidandrewfrey/pen/wWLxNq

davidandrewfrey:
⭐ ⭐ ⭐ ⭐ ⭐ 
Thank you for making the updates

PR #605, opened 2016-08-26
invisible lists 2: invisible boogaloo

nelsonpecora:
* adds generic invisible list support
* uses property `invisible: true` in `_componentList`
* components name shows head lists, then invisible lists
* updated pane tab margins to fit more of them
* added `overflow-x: scroll` to pane tabs

screen shot 2016-08-26 at 2 49 20 pm

### DIFF
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -164,6 +164,39 @@ sidebar:
     page: true
 ```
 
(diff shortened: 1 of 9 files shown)


jonwinton:
👍

PR #606, opened 2016-08-26
Component Search Close Pane On Selection

jonwinton:
Like the title says, just closes the pane when a component is selected so a user can interact with it immediately.


nelsonpecora:
👍🏼

PR #607, opened 2016-08-29
Fix: Add component sometimes closing tabs

jonwinton:
Updated to call `pane.close()` instead of `window.close()` which might close a browser tab.


dmabuada:
+1

PR #611, opened 2016-08-31
fuzzy adding components

nelsonpecora:
* [trello ticket](https://trello.com/c/n7cR8jyg/27-add-one-off-components)
* allows adding ANY available component if allowd by a component list
* depends on amphora PR https://github.com/nymag/amphora/pull/305
* adds `all-components` service to make the code DRYer

### DIFF
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -188,7 +188,7 @@ It's useful to have a component list for various components that might not have 
 
 ```yaml
 footerMetadata:
-  _component:
+  _componentList:
     include:
       - pixel-tracker
       - gpt-script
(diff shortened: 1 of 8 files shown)


behaviors/README.md
```diff
@@ -197,6 +197,28 @@ footerMetadata:
 
 Like the Page Area property, this only works in a layout's component lists.
 
+#### Fuzzy Component Lists and Properties
+
+By default, component lists and properties allow strict whitelisting and blacklisting with `include` and `exclude`. This is appropriate for most situations, but there are certain scenarios where you might want to be less strict. For example, the body of an article might be paragraphs and images 99% of the time, but you still want to allow your interactives team to create one-off components and add them to articles.
+
+You can accomplish this by adding `fuzzy: true` to your component list / property. This will add a "View all components" button to the bottom of the component list in the add-components pane. When a user clicks the button, it'll open a different add-components pane with a list of all available components.
```
cruzanmo: This sounds like a way to hide the component as an option? Maybe `hidden: true`?

nelsonpecora: No, it makes the white/black list less strict by allowing users to (with an extra action) add _any_ component available in the clay instance. This allows for things like one-off interactives in articles (without having to manually to api calls against the server to add them)

cruzanmo: Oh, I understand, that makes sense.

What about calling it `allowAll` or `allowAny`?

nelsonpecora: we were debating that, but the discussion naturally settled on using `fuzzy` to talk about these kinds of lists

cruzanmo: hmm, not my first choice, but if the team thinks it makes sense, that's cool.

jonwinton:
👍

PR #612, opened 2016-08-31
Time input fix

jonwinton:
**Bug:** time inputs would sometimes receive an `undefined` date value which would then clear out the date when users would try to change the input values.

**Fix:** check to see if there is a value for the input at instantiation of the input and and if not create use an empty string as the value.


nelsonpecora:
👍

PR #616, opened 2016-09-07
Component Ref In form:open event

jonwinton:
Use Case: You want a plugin to be aware of form openings, but only act if a specific component's form is opening. Passing a `ref` provides a way to select the component and determine which component it is.


byronhulcher:
👍

nelsonpecora:
👍

cruzanmo:
+1

PR #618, opened 2016-09-08
Adding packages back in for karma testing

jonwinton:
When converting to Webpack I removed packages that I thought were only used in the Gulp compile process. Because the packages were already installed in my environment the local testing worked for me. For some reason the CircleCI tests also passed without those packages. But when trying to publish to NPM the tests didn't work. This should fix that issue.


nelsonpecora:
👍

PR #620, opened 2016-09-12
fixed page-specific component reorder issue

nelsonpecora:
* [trello ticket](https://trello.com/c/ORBbUzj1/1169-bug-article-jumping-into-layout)
* fixed issue where reordering page-specific components in a layout would save to the layout data rather than the page data


byronhulcher:
👍

PR #622, opened 2016-09-13
fixed schedule form in firefox

nelsonpecora:
this fixes the schedule form in firefox, and right-aligns the time picker so it flows correctly.

screen shot 2016-09-13 at 10 54 27 am


byronhulcher:
👍

PR #624, opened 2016-09-13
convert all headers to strong tags

nelsonpecora:
we had some issues with people accidentally pasting in headers into paragraphs, which totally breaks the editing experience (since `<h2>` tags cannot live inside `<p>` tags). This converts any pasted in header to bold text, rather than simply preventing the paste.

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -14,6 +14,17 @@ var _ = require('lodash'),
   site = require('../services/site'),
   refAttr = references.referenceAttribute;
 
+// pass config actions to text-model
+model.updateSameAs({
+  // all headings should be converted to bold text
+  H1: 'STRONG',
+  H2: 'STRONG',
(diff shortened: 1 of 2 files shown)


jonwinton:
👍

behaviors/wysiwyg.js
```diff
@@ -14,6 +14,17 @@ var _ = require('lodash'),
   site = require('../services/site'),
   refAttr = references.referenceAttribute;
 
+// pass config actions to text-model
+model.updateSameAs({
```
byronhulcher: Stuff like this makes me miss python
`{"H%s"%(index): "STRONG" for index in range(1,7)}`

nelsonpecora: `_.each([1, 2, 3, 4, 5, 6], (i) => obj[`H${i}`] = 'STRONG';)`

PR #625, opened 2016-09-13
Filtered List Item Title Width Bug

jonwinton:
screen shot 2016-09-13 at 1 54 28 pm

Small issue where the content of a filtered list item title grows past the width of the pane based on its content. The `flex-shrink` needs to be set to `1` so that it doesn't grow too wide, and since no other items in the flexed container have a `flex-grow` of `1` or more, the title will always grow to fill the space.

### DIFF
--- a/styleguide/filterable-list.scss
+++ b/styleguide/filterable-list.scss
@@ -68,7 +68,7 @@
 .dragula-list-item .filtered-item-title {
   @include primary-text();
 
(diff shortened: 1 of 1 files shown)


jonwinton:
(There should be settings/delete buttons to the right of the title)

nelsonpecora:
ok 👍

PR #628, opened 2016-09-19
Get Default Data By Component Name

jonwinton:
Adding functionality to retrieve a component's default data with just the component's name.


nelsonpecora:
👍

PR #633, opened 2016-09-20
simple-list fix 

amycheng:
fixes https://github.com/nymag/clay-kiln/issues/632


behaviors/simple-list.js
```diff
@@ -146,6 +146,8 @@ module.exports = function (result, args) {
       // returns false if repitition is allowed
       // returns false if repitition is disallowed and items don't repeat
       function hasRepeatedValue(value, data) {
+        //if data is undefined make it an empty array
+        var data = !data ? [] : data;
```
nelsonpecora: this should probably be done above in the `bind` function (it's where the data initializes). Also, keep in mind our linter doesn't like multiple `var` declarations 😄

amycheng: implemented

fraddski:
:+1:

PR #640, opened 2016-09-26
Expose Behavior Add Function In Services

jonwinton:
**Purpose:** Allow behaviors to be registered through Kiln's normal behavior add process

**Advantage:** Makes it so arguments can be passed to custom behaviors.


nelsonpecora:
👍  but maybe it should just expose `module.exports` like other services? Your call.

PR #642, opened 2016-09-26
new selectors

nelsonpecora:
* [trello ticket](https://trello.com/c/34apsQdd/47-selectors)
* removed parent button
* added prev/next buttons
* added this page / multiple pages icon
* added ↑/↓ keyboard navigation when a component is selected (but no form is open)
* removed unselecting when closing forms (press <kbd>esc</kbd> again to unselect), allowing better navigation
* moved `visible-components` into a service
* added document padding if selectors would overflow top/bottom of the document
* removed janky selection of `layout` itself
* removed parent selector
* fancy new selector styles (background fade, consistent menus, animated hide/show)

499cc462-71c0-4485-ba15-0b7d8209dd7e-1015-00003e40b88907ec

### DIFF
--- a/media/parent.svg
+++ b/media/parent.svg
@@ -1,1 +1,0 @@
-<svg width="18" height="16" viewBox="0 0 18 16" xmlns="http://www.w3.org/2000/svg"><path fill="#727272" d="M12.706 6.353l-1.504 1.503-3.79-3.8v9.709H18v2.117H5.294V4.055l-3.79 3.801L0 6.353 6.353 0z" fill-rule="evenodd"/></svg>
--- a/services/components/select.test.js
+++ b/services/components/select.test.js
@@ -1,3 +1,4 @@
+/* eslint max-nested-callbacks: [1, 5] */
 var dirname = __dirname.split('/').pop(),
   filename = __filename.split('/').pop().split('.').shift(),
   references = require('../references'),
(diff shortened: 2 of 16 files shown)


services/components/select.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint max-nested-callbacks: [1, 5] */
```
byronhulcher: What's this about? Looks like you only removed code in these tests

styleguide/component-select.scss
```diff
@@ -3,179 +3,190 @@
...
+    top: auto;
+  }
 }
 
-// bottom menu is more offset, since it concerns actions that don't
-// affect the component itself (but rather things around it, e.g. adding new components)
 .component-selector-bottom {
-  border-color: $selector-border;
-  top: 100%;
+  top: calc(100% - 96px); // account for toolbar
```
gloddy: Is 96px a desktop+OS specific number? Anything we can tie this to that is a little more cross-platform?

nelsonpecora: Nope, it's based on the kiln toolbar height

services/components/select.test.js
nelsonpecora: Yeah, but deleting the code exposed a linting issue. we massage the max-nested-callbacks a little bit in a few of our tests

styleguide/component-select.scss
gloddy: Ah. Never mind then.

gloddy:
👍

media/parent.svg
```diff
@@ -1 +0,0 @@
-<svg width="18" height="16" viewBox="0 0 18 16" xmlns="http://www.w3.org/2000/svg"><path fill="#727272" d="M12.706 6.353l-1.504 1.503-3.79-3.8v9.709H18v2.117H5.294V4.055l-3.79 3.801L0 6.353 6.353 0z" fill-rule="evenodd"/></svg>
```
jonwinton: Hmm so we're getting rid of the parent button entirely? Does this mean we're relying on the arrow navigation to get there and can't jump quickly?

nelsonpecora: Correct.

nelsonpecora: Personally I really like the parent button, but it wouldn't fit well with the new navigation scheme.

nelsonpecora:
@jonwinton with the changes you made to `clay-space-edit`, is this ok to merge now?

jonwinton:
👍  So pretty


But let's coordinate release per our sidebar.

PR #644, opened 2016-09-30
Pane Events

jonwinton:
- Added event for pane openings (`pane:open`) and closes (`pane:close`)
- Added an event for when an invisible component list is generated for the Component Pane


nelsonpecora:
👍

PR #645, opened 2016-09-30
stop propagation when clicking overlays

nelsonpecora:
* stop propagation when clicking overlay itself, since we call `unfocus()`
* stop propagation when clicking form inside an overlay, since we don't want to unselect the current component


jonwinton:
👍

PR #646, opened 2016-09-30
updated keyboard shortcuts

nelsonpecora:
* <kbd>esc</kbd> closes all panes, then closes any open form, then unselects components
* prevent up/down arrows in filterable list from navigating components
* removed unused `add-component-pane` controller
* [trello ticket](https://trello.com/c/eqoMzZE5/52-esc-should-close-all-panes)


jonwinton:
👍  I like that there's more red than green and things work more properly

PR #647, opened 2016-09-30
call forms directly for a variety of actions

nelsonpecora:
There have been a few issues with things using `focus.unfocus()` and `focus.hasCurrentFocus()` (which deals with inline and overlay forms, but not forms that are opened programmatically by anything other than a user clicking a `data-editable` element in a component) rather than calling the `forms` service directly. This included:

* <kbd>esc</kbd> not working on settings forms
* programmatic forms not being closed when opening the new page, preview, or publish pane
* "Are you sure you want to leave this page?" confirmation not being served
* "Are you sure you want to create a new page?" confirmation not being served

### DIFF
--- a/controllers/kiln-toolbar.js
+++ b/controllers/kiln-toolbar.js
@@ -1,7 +1,7 @@
 const dom = require('@nymag/dom'),
   keyCode = require('keycode'),
   events = require('../services/events'),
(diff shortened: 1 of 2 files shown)


jonwinton:
Woo 👍

PR #650, opened 2016-10-03
selector feedback

nelsonpecora:
* slightly less fade on background components
* on larger screens, only show selectors on hover
* (todo: [when firefox supports pointer media queries](https://bugzilla.mozilla.org/show_bug.cgi?id=1035774), use that instead of just viewport width)

b8edbab3-9447-4304-82d8-206ee0d554cf-1015-0000b52c4c3ce997


cruzanmo:
+1

gloddy:
👍 Worth a shot.

PR #651, opened 2016-10-04
Fix to Simple List Behavior to account for undefined data

amycheng:
this fixes a bug where if data for a Clay field that uses the `simplelist` behavior is undefined it throws two errors:

* `e.map() is in undefined`
* `... .toLowerCase is undefined`


nelsonpecora:
👍

PR #652, opened 2016-10-04
the correct(???) way to account for undefined data being passed into the simple-list behavior

amycheng:
so the previous fix was overwriting ALL data causing previously entered data (like authors) to be deleted. This fix is another way of checking whether or not the data is defined. See my comments for more context.

### DIFF
--- a/behaviors/simple-list.js
+++ b/behaviors/simple-list.js
@@ -139,11 +139,15 @@ module.exports = function (result, args) {
       // this is called when the binder initializes
       var addEl = dom.find(boundEl, '.simple-list-add'),
         allowRepeat = !!(addEl.getAttribute('data-allow-repeat') === 'true'),
-        // if data is undefined make it an empty array
-        data = data || [],
         observer = this.observer;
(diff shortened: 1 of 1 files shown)


behaviors/simple-list.js
```diff
@@ -139,11 +139,15 @@ module.exports = function (result, args) {
       // this is called when the binder initializes
       var addEl = dom.find(boundEl, '.simple-list-add'),
         allowRepeat = !!(addEl.getAttribute('data-allow-repeat') === 'true'),
-        // if data is undefined make it an empty array
-        data = data || [],
```
amycheng: this conditional was always returning `undefined`.

behaviors/simple-list.js
```diff
@@ -139,11 +139,15 @@ module.exports = function (result, args) {
...
         observer = this.observer;
 
-      observer.setValue(data);
```
amycheng: this line was overwriting all data with a blank array.

PR #660, opened 2016-10-11
change selector fade from 60% to 40%

nelsonpecora:
Making the background slightly more opaque based on user feedback.

screen shot 2016-10-11 at 11 22 21 am


amycheng:
👍

PR #661, opened 2016-10-11
fix some selector hover issues

nelsonpecora:
* [trello ticket](https://trello.com/c/9VWRves3/54-selector-hover-updates)
* use `hover` media query (fixes tablet hover issues)
* add firefox workaround (firefox doesn't support `hover` media query yet)
* display selectors for a second when initially selecting component (for better navigation)

c0453853-047f-4a95-9430-1ec0406287a5-1015-000151661a900be1

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,4 +1,5 @@
 'use strict'; // eslint-disable-line
+
 // note: use strict above is applied to the whole browserified doc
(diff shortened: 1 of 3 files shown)


gloddy:
👍 One step closer to awesomepants.

PR #666, opened 2016-10-20
deep linking

nelsonpecora:
* [trello ticket](https://trello.com/c/AH8tEwYY/51-deep-linking)
* hash will change when opening forms
* hash will change when closing forms
* loading the page w/ a hash will automatically open form
* clicking validation links will open relevant form

Note: The validation links will be applied if the particular validator passes in a `fieldName` and the component's `data`. Certain validators (`max-instances`, `component-string-pairs` don't really apply to individual fields, thus they won't generate links in the validation tab)

### DIFF
--- a/services/pane/publish.js
+++ b/services/pane/publish.js
@@ -9,6 +9,7 @@ const _ = require('lodash'),
   pane = require('./'),
   datepicker = require('../field-helpers/datepicker'),
   publishPaneController = require('../../controllers/publish-pane'),
+  link = require('../deep-link'),
   kilnHideClass = 'kiln-hide';
 
 /**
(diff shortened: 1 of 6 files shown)


services/pane/publish.js
```diff
@@ -88,7 +107,17 @@ function addErrorsOrWarnings(errors, modifier) {
 
     // add each place where the error/warning occurs
     _.each(error.errors, function (item) {
-      var itemEl = dom.create(`<li><span class="error-label">${item.label}</span>${addPreview(item.preview)}</li>`);
+      let itemEl;
+
+      if (item.fieldName && item.data) {
+        // add a link to open the specific form to update the data
+        itemEl = dom.create(`<li data-error-ref="${item.ref}" data-error-path="${link.getPathFromField(item.fieldName, item.data)}"><span class="error-label error-label-link">${item.label}</span>${addPreview(item.preview)}</li>`);
```
jonwinton: Why not use an anchor tag for the link? I know you need multiple pieces of data (`error-path` and `error-ref`) and so one `href` value wouldn't work, but it should be a link semantically right?

PR #671, opened 2016-10-31
view and edit users

nelsonpecora:
* [trello ticket](https://trello.com/c/eOGXMKjS/57-people-pane)
* basic clay menu (w/ settings tab + people button)
* people pane w/ Me, People, and Log Out
* you can edit your name and title
* you can see your image, username, and provider
* you can log out of clay
* you can see a list of users on your current site (when we have a ui for adding users, this should probably grab users from all sites?)
* added ability for panes to open on the left or right side of the screen
* updated components pane to open on the left side of the screen


gloddy:
@nelsonpecora We have two possible panes competing for the title of `People`, so it's probably worthwhile to pick the winner now.
The two are:
- The `People` pane that shows the list of everyone in a Clay install
- The `People` pane that shows people associated with a specific page

I'm currently learning towards the first being called `Directory` or something similar.

PR #677, opened 2016-11-04
Move `toolbar-layer` above `overlay-layer`

byronhulcher:
Otherwise elements in the `kiln-toolbar-pane` are beneath elements in the `editor-overlay`

One example of this is a field edited in `_display: overlay` that opens a mediaplay image picker, which renders within the `kiln-toolbar-pane`

### DIFF
--- a/styleguide/_layers.scss
+++ b/styleguide/_layers.scss
@@ -6,12 +6,12 @@
   z-index: 1000;
 }
 
-@mixin toolbar-layer {
-  z-index: 1010;
-}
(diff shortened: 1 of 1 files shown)


nelsonpecora:
👍

cruzanmo:
Looks like it works. Odd that the code coverage decreased.

nelsonpecora:
Kiln coverage is a bit wonky right now. Don't worry too much about it 😄

PR #695, opened 2016-11-28
aggressively cache schemas after they're loaded

nelsonpecora:
* cache schemas in client-side memory when they're loaded
* note: all schemas for components _on the page_ when the page is loaded will be cached (they're called so that decorators can run)
* this should speed up component creation and saving


cruzanmo:
+1

PR #701, opened 2016-12-05
redesigned status messages

nelsonpecora:
* [trello ticket](https://trello.com/c/EYkRev1i/62-clay-status-messages)
* new semantic colors (slightly more pastel than before)
* larger status message (now that they aren't permanent)
* removed permanent status messages (schedule and publish)
* merged `page` and `layout` colors into `save` (blue)
* created `progress.error` convenience method to pass errors through
* note: `offline` status message will be permanent until you reconnect
* passing in error messages directly to error status messages now

screen shot 2016-12-05 at 4 13 42 pm

screen shot 2016-12-05 at 4 14 31 pm

screen shot 2016-12-05 at 4 15 43 pm

### DIFF
--- a/controllers/kiln-toolbar-edit.js
+++ b/controllers/kiln-toolbar-edit.js
@@ -25,10 +25,7 @@ let EditorToolbar;
 function createPage() {
   return forms.close()
     .then(openNewPage)
-    .catch(function () {
-      progress.done('error');
-      progress.open('error', 'Issue with opening page options.', true);
-    });
+    .catch(progress.error('Error opening new pages list'));
 }
 
 /**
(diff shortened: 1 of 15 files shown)


controllers/kiln-toolbar-edit.js
```diff
@@ -128,16 +121,12 @@ EditorToolbar.prototype = {
         return validation.validate(rules).then(function (results) {
           if (results.errors.length) {
             progress.done('error');
+            progress.open('error', 'Error opening publish', results.errors.join(', '));
           }
 
-          openPublish(results);
+          return openPublish(results);
```
nelsonpecora: return this promise, so any issues will flow down to the `.catch()` below

gloddy:
browserstack!!

nelsonpecora:
more like...

( •_•)>⌐■-■

BOWSER stack

(⌐■_■)

PR #707, opened 2016-12-15
fix issue replacing components with components after them

nelsonpecora:
thanks @byronhulcher!


services/edit/index.js
```diff
@@ -563,13 +563,15 @@ function addToParentList(opts) {
  * @param {object} opts
  * @param {array} opts.refs
  * @param {string} [opts.prevRef]     The ref of the item to insert after.
+ * @param {number} [insertIndex] actual index to insert first item into (allows adding multiple components above existing components)
```
byronhulcher: plz add `opts.` to this

PR #708, opened 2016-12-16
Exposes layout object to validators

cperryk:
This change passes the layout object to validation functions so we can develop layout-specific rules. (Relevant [Trello](https://trello.com/c/srUYDxPl/1316-rendition-layout-pre-publish-validation))

### DIFF
--- a/services/publish-validation.js
+++ b/services/publish-validation.js
@@ -94,12 +94,17 @@ function mapRules(rules, state) {
  */
 function validate(rules) {
   var refs = getLatestRefMap(),
-    components = getComponentMap(refs);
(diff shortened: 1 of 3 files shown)


jonwinton:
This would also require an update of documentation: https://github.com/nymag/clay-kiln/blob/master/validators/README.md

nelsonpecora:
👍🏼 from sunny CA

nelsonpecora:
note: in kiln v1.0, we're passing the full state object, so you can grab the layout uri (`state.page.data.layout`) or data (`state.components[layoutURI]`) in the validator yourself.

PR #713, opened 2016-12-27
Adding back `selected-parent` class

jonwinton:
As components grow more complex, sometimes we need to write styles specific to edit mode. Sometimes in this situation we need awareness of when a component's parent is selected. There used to be a class for this, but in some updates it seems to have been removed. This adds that class back in. 

**Example**
The `in-article-slideshow` component has a component-list of slides. If we want these all to be editable we need to show them all on the page, which makes the page very long _and_ makes it so that the component cannot be dragged to different locations on the page. By having selected mode styles only we can maintain it's proper appearance and only expand it when we want to edit the component.

**When not editing**
screen shot 2016-12-27 at 10 33 03 am

**When Editing**
screen shot 2016-12-27 at 10 33 20 am

### DIFF
--- a/services/components/select.js
+++ b/services/components/select.js
@@ -16,7 +16,8 @@ var _ = require('lodash'),
   visibleComponents = require('./visible-components'),
   hidden = 'kiln-hide',
   selectorHeight = 56, // selector menus are 48px tall, offset is 8px
-  currentSelected;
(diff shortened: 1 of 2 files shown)


dmabuada:
looks ok. +1

PR #718, opened 2017-01-03
paste fixes

nelsonpecora:
* [trello ticket](https://trello.com/c/2wVlnmNv/77-bug-pasting-rules-for-multi-line-paragraphs-don-t-work-properly)
* **BREAKING CHANGE:** adds `matchLink` to positively match urls in links (rather than assuming them by default, which broke links inside paragraphs when the WHOLE paragraph was enclosed in a single link)

Updates the component splitting regex:

* split on double `<br>` tags (rather than just one)
* split on `<br>\n`, `\n<br>`, and `\n\n`
* allow whitespace before, after, and between those double line breaks

Cleans the matches:

* removes `<p><br>` at the beginning of strings, not just `<p>` or `<br>`
* removes matches that simply contain a closing tag (it's a google docs edge case)

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -16,7 +16,7 @@ var _ = require('lodash'),
 
 // pass config actions to text-model
 model.updateSameAs({
-  // all headings should be converted to bold text
+  // all headings (inside the current component) should be converted to bold text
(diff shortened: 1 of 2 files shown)


dmabuada:
this looks ok 👍

PR #732, opened 2017-01-17
Handlebars

nelsonpecora:
* use handlebars template for kiln
* reference `@root.locals` property

Notes:

* svgs are read from the `media` folder, rather than the (assumed) `../../public/media/components/clay-kiln/` folder. They should be optimized
* this only converts the template to handlebars. it doesn't add client-side handlebars rendering of components or anything (that will happen later)
* assets grabbed client-side (e.g. the formatting icons in the wysiwyg toolbar) are still being grabbed from the `public` folder


template.handlebars
```diff
@@ -1,23 +1,23 @@
-{%- if locals.user -%}
-  {# if a user is logged in, we display either the view-mode or edit-mode toolbar, and associated scripts, styles, and templates #}
-  <div data-uri="{{ _ref or _self }}" class="clay-kiln">
+{{#if locals.user}}
```
amycheng: question: why is `@root` not used here? is it because this template is IN the root?

template.handlebars
```diff
@@ -1,23 +1,23 @@
...
-    {# show different styles in view/edit mode #}
+    {{! show different styles in view/edit mode }}
     <style>
-      {%- if locals.edit -%}
-        {{ ('node_modules/clay-kiln/dist/clay-kiln-edit.css')|includeFile|safe }}
-      {%- else -%}
-        {{ ('node_modules/clay-kiln/dist/clay-kiln-view.css')|includeFile|safe }}
-      {%- endif -%}
+      {{#if locals.edit}}
+        {{{ read 'node_modules/clay-kiln/dist/clay-kiln-edit.css' }}}
```
amycheng: question: is `read` the same as nunjucks' `include`?

nelsonpecora: pretty much, though you need to use triple-brackets to escape the contents

PR #734, opened 2017-01-17
better error handling

nelsonpecora:
better error handling for certain situations, especially when pasting in multiple components.

if one of the pasted components errors out (e.g. gives a 502 or 500 error), this will error out the whole operation rather than silently failing to paste

### DIFF
--- a/services/edit/index.js
+++ b/services/edit/index.js
@@ -255,7 +255,7 @@ function createPage(pageType) {
   return cache.getDataOnly(newPageUri).then(function (data) {
     return db.create(prefix + pagesRoute, _.omit(data, '_ref')).then(function (res) {
       return getNewPageUrl(res[refProp]);
-    });
+    }).catch(progress.error('Error creating page'));
   });
 }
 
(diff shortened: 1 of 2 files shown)


services/progress.js
```diff
@@ -130,7 +130,7 @@ function error(message) {
     open('error', `${message}: ${e.message}`);
     // note: "Failed to fetch" is a common error message which means an API call failed.
     // check the console to see which calls are giving issues.
-    return e;
+    return Promise.reject(e); // reject error rather than returning it
```
nelsonpecora: this is the important bit. we don't want `.then()` to run if we're explicitly catching an error

services/edit/index.js
```diff
@@ -371,7 +371,7 @@ function createComponent(name, data) {
     } else {
       return cache.createThrough(instance, defaultData);
     }
-  });
+  }).catch(progress.error('Error creating component'));
```
amycheng: is it possible to pass in a component name or ref?

nelsonpecora: oh good point! adding

PR #736, opened 2017-01-18
preload schemas server-side

nelsonpecora:
* adds schemas of components (currently on the page) to inline js server-side
* uses that exposed schema cache when grabbing new schemas
* note: once we have versioned schemas, we can store this cache in localStorage and get a pretty big performance improvement
* this requires an update to amphora

### DIFF
--- a/template.handlebars
+++ b/template.handlebars
@@ -20,7 +20,7 @@
         window.kiln.services.schemaCache = window.kiln.services.schemaCache || {};
 
         {{#each @root._componentSchemas}}
-          window.kiln.services.schemaCache['{{ this.name }}'] = {{{ stringify this.schema }}};
+          window.kiln.services.schemaCache['{{ this.name }}'] = {{{ stringify (yaml (read this.schema)) }}};
         {{/each}}
       </script>
     {{/if}}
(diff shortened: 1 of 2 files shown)


template.handlebars
```diff
@@ -11,6 +11,22 @@
...
+        // note: this must declare all the levels, since kiln js hasn't loaded yet
+        window.kiln = {
+          services: {
+            schemaCache: {
+              {{#each @root._componentSchemas}}
+                '{{ this.name }}': {{{ stringify this.schema }}},
+              {{/each}}
+            }
+          }
+        }
```
cruzanmo: This works, but seems fragile. Would it be possible to use `_.set` or the slightly longer `window.kiln = window.kiln || {}; window.kiln.services = window.kiln.services || {}; ...`, or some other way to ensure nothing is overwritten?

nelsonpecora: Hmm. It always happens before kiln js loads, but plugins and such might add stuff before it. good point

nelsonpecora:
Updated to use file paths. Depends on having a `yaml` helper in `nymag-handlebars`

nelsonpecora:
Dependencies for this are merged in

nelsonpecora:
This can be merged in and released before the amphora changes are done (tested with amphora 2.9.1)

PR #738, opened 2017-01-19
Replace all non word chars for magic button URL

fraddski:
When formatting a URL for the magic button, all non-word characters should be replaced by hyphens.


nelsonpecora:
👍

PR #740, opened 2017-01-20
add queue

nelsonpecora:
* all PUTs (not POST or GET) will be synchronous, preventing people from getting into bad states because of server errors
* form closing/unfocus/unselect doesn't happen until stuff saves, preventing weird in-between states
* you can drag+drop multiple items as API calls are queued, but they'll retain their unsaved (blue) highlight until they save (and now they're guaranteed to save _in order_)
* when the page loads, there's now a progress bar as well as the "Loading" message
* when navigating away from the page, the queue will be checked (as well as any open forms) and an confirm will pop up if there are pending API calls


services/edit/queue.js
```diff
@@ -0,0 +1,55 @@
...
+ * but promises may contain async children (e.g. Promise.all([a, few, api, calls]))
+ * @param {Function} fn
+ * @param {Array}   args
+ * @param {string} [progressType] 'publish', 'schedule', add this if you need to end with a different progress bar color
+ * @returns {Promise}
+ */
+function add(fn, args, progressType) {
+  const cacheHash = fn.name + JSON.stringify(args);
+
+  let newPromise;
```
jonwinton: Why is this a `let` instead of a `var`? I don't _really_ care, but we're not dealing with block scope here, so I just wanted to make sure there wasn't a different reasoning

nelsonpecora: just favoring const and let rather than var 😄

jonwinton: Haha can't say I necessarily agree with favoring `let`, but I don't feel strongly enough about it to raise a flag. Looks good 👍

PR #741, opened 2017-01-24
HUGE kiln speed improvements

nelsonpecora:
* fixed syntax error in queue which wasn't properly caching promises
* _all_ component saves/gets now return without `combined-<site>.css` (the fonts), which means they're exponentially smaller

### DIFF
--- a/services/edit/db.js
+++ b/services/edit/db.js
@@ -5,7 +5,7 @@ var _ = require('lodash'),
   site = require('./../site'),
   rest = require('./rest'), // todo: use native fetch when safari supports it
   extHtml = '.html',
-  editMode = '?edit=true',
+  editMode = '?edit=true&sitecss=false',
   // when we ask for updated component html,
   // make sure the server knows we're in edit mode
   componentRoute = '/components/',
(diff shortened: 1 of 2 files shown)


services/edit/db.js
```diff
@@ -5,7 +5,7 @@ var _ = require('lodash'),
   site = require('./../site'),
   rest = require('./rest'), // todo: use native fetch when safari supports it
   extHtml = '.html',
-  editMode = '?edit=true',
+  editMode = '?edit=true&sitecss=false',
```
nelsonpecora: this part will be enabled by https://github.com/nymag/sites/pull/2318

services/edit/queue.js
```diff
@@ -34,7 +34,7 @@ function add(fn, args, progressType) {
 
   // every time we add a function to the queue, check to see if it's already added
   if (queueCache[cacheHash]) {
-    return Promise.resolve(queueCache[cacheHash]());
+    return Promise.resolve(queueCache[cacheHash]);
```
nelsonpecora: this means "resolve the promise that's currently pending", rather than the old one which, by calling the function again, actually resolved a new promise

PR #752, opened 2017-02-02
client-side models and templates

nelsonpecora:
* adds support for `model.js` and handlebars templates
* relies on https://github.com/nymag/amphora/pull/372 but isn't blocked by it
* will use `model.save` and `model.render` if they're available
* assumes components with `model.js` have client-side templates

NOTES:

* doesn't handle errors well. it'll throw them to the user, but won't revert nicely if API calls fail (requires redux)
* doesn't have all `locals` that server-side has (we'll add more as we move components over)
* doesn't create + rerender components, only does client-side stuff when saving already-created components


edit.js
```diff
@@ -16,7 +16,9 @@ var references = require('./services/references'),
   plugins = require('./services/plugins'),
   eventify = require('eventify'),
   link = require('./services/deep-link'),
-  connectionLostMessage = 'Connection Lost. Changes will <strong>NOT</strong> be saved.';
+  connectionLostMessage = 'Connection Lost. Changes will <strong>NOT</strong> be saved.',
+  behaviorReq = require.context('./behaviors', false, /^\.\/[\w\-]+\.js$/), // no *.test.js
+  decoratorReq = require.context('./decorators', false, /^\.\/[\w\-]+\.js$/); // no *.test.js
```
nelsonpecora: use `require.context` to automatically find new behaviors and decorators as we add them, rather than having to manually specify them below (this is a feature webpack gives us that browserify did not)

services/edit/db.js
```diff
@@ -300,11 +300,17 @@ function getHTMLWithQuery(uri, queries) {
...
+ * @param {boolean} [componenthooks] optional, true by default
  * @returns {Promise}
  */
-function save(uri, data) {
+function save(uri, data, componenthooks) {
   assertUri(uri);
 
+  // set `false` to disable component models from firing their hooks
+  if (componenthooks === false) {
+    uri = uri + '?componenthooks=false';
```
nelsonpecora: I'll pull this up when we start using it for `db.create`

PR #755, opened 2017-02-03
add partials for all component templates

nelsonpecora:
* register partials using the precompiled templates on load
* this allows components to embed other components completely client-side


karma.conf.js
```diff
@@ -18,7 +18,6 @@ const files = [
           loader: 'babel-loader',
           options: {
             presets: ['es2015'],
-            cacheDirectory: true,
```
nelsonpecora: this messes with coverage on circle

PR #763, opened 2017-02-07
pasting improvements

nelsonpecora:
* REAL error messages when bad things are pasted. also doesn't allow the bad pasted content to be saved into the system.
* parses out text before and after blockquotes (as separate paragraphs)
* removes deeply-nested `<p>` and `<div>` tags
* removes `line separator` and `paragraph-separator` characters
* converts tabs to spaces
* converts nonbreaking spaces to spaces
* converts newlines that AREN'T EXPLICITLY BETWEEN SENTENCES to spaces
* (newlines between sentences are converted to `<br>` tags)
* trims the text after applying all those transforms

Also removes the uglify plugin, as the code is minified by default on release builds but the ulgify plugin messes with watching the code when developing.

* fixes [issue with blockquotes](https://trello.com/c/SIJ1VZXK/1382-clay-when-pasting-text-in-clay-text-doesn-t-paste-after-a-blockquote-tag)
* fixes [other paste issues](https://trello.com/c/0FvAW63k/94-bug-nested-tags-create-uneditable-paragraphs-in-clay)

### DIFF
--- a/behaviors/wysiwyg.js
+++ b/behaviors/wysiwyg.js
@@ -54,7 +54,7 @@ function selectAfter(node) {
  * @returns {array}
  */
 function splitParagraphs(str) {
-  // </p>, </div>, </h1> through </h9>, </blockquote>, or two (interchangeable) <br> or newlines
+  // </p>, </div>, </h1> through </h9>, or two (interchangeable) <br> or newlines
   // note: <br> tags may contain closing slashes, and there may be spaces around stuff
   // note: split on both </blockquote> and <blockquote>, since there may be text before/after the quote
   let paragraphs = _.map(str.split(/(?:<\/(?:p|div|h[1-9])>|(?:\s?<br(?:\s?\/)?>\s?|\s?\n\s?){2})/ig), s => s.trim());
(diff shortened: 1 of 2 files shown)


behaviors/wysiwyg.js
```diff
@@ -71,19 +93,41 @@ function splitParagraphs(str) {
...
+    let cleanStr, matchedRule, matchedObj, matchedValue;
+
     // remove extraneous opening <p>, <div>, and <br> tags
     // note: some google docs pastes might have `<p><br>`
-    var cleanStr = str.replace(/^\s?<(?:p><br|p|div|br)(?:\s?\/)?>\s?/ig, ''),
-      matchedRule = _.find(rules, function matchRule(rule) {
-        return rule.match.exec(cleanStr);
-      }),
-      matchedObj, matchedValue;
+    cleanStr = str.replace(/^\s?<(?:p><br|p|div|br)(?:.*?)>\s?/ig, '');
```
amycheng: I know this is merged already but what do you think about adding all these regexs to an array and then looping through that array?

So something like:

```
let bad = [
/<(?:p|div).*?>/ig, / // remove any other <p> or <div> tags, because you cannot put block-level tags inside paragraphs
/(\u2028|\u2029)/g //
];

...

_.forEach(...{
  cleanStr = cleanStr.replace(bad[i], ' ');
})

cleanStr = cleanStr.replace(/\.\n[A-Z0-9]/g, '<br>');
cleanStr = cleanStr.trim();
```

amycheng:
💯  code comments

behaviors/wysiwyg.js
nelsonpecora: Hmm! Let me take a look at the v1.0 workflow, and see if that still applies (I think it does)

PR #803, opened 2017-04-06
Cache lists in Vuex store

mheiber:
Instead of requesting list data from Amphora each time an
autocomplete behavior needs it, check to see if it is in the Vuex
store, and get it from the Vuex store if available.
If it's not available, dispatch an action which causes the lists
to be fetched.

Track pending/success/failure state of the lists request. Not
doing anything with it yet, but it's there if we want to reflect
this state in the UI or use Vue devtools to see what's happening.

Doesn't buy us much beyond cleanliness, since the browser will
probably cache the results of the Amphora calls anyway, but
it's a little cleaner to keep stuff in the store.

In this commit, I also added the first module to the Vuex store.
Seems like an easier way to divide up the Vuex stuff, and is
similar to `combineReducers` in Redux.

I deleted the existing `lists` stuff in the store so as not to
have name conflicts. As far as I can tell, that stuff wasn't used.

This is prep work for a change I hope to make in another PR: adding an argument to the `simple-list` and `test` behaviors so that we can validate that the user's entry is a member of the given list.

### DIFF
--- a/behaviors/autocomplete.vue
+++ b/behaviors/autocomplete.vue
@@ -33,18 +33,6 @@
 <script>
   import _ from 'lodash';
   import item from './autocomplete-item.vue';
-  import { getJSON } from '../lib/core-data/api';
-  import { uriToUrl } from '../lib/utils/urls';
-
-  const textProp = 'text';
-
-  function flattenText(items) {
-    var pluckedText = _.compact(_.map(items, textProp)),
-      hasTextProp = _.isString(_.head(pluckedText));
-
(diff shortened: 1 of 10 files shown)


nelsonpecora:
oooh nice! yeah, jon and I were debating whether we should preload the data for these, and I think your implementation works. let me test it tomorrow

mheiber:
Thanks for reviewing, @nelsonpecora , I'll add tests.

mheiber:
@nelsonpecora I added tests, good call 👍

PR #804, opened 2017-04-12
use date-fns instead of moment

nelsonpecora:
* removed extraneous checks from some behaviors (vue silently eats those errors)
* convert `text` behavior and publish panes to use `date-fns` instead of moment
* use smaller `nymag-handlebars` version (1.22.0)
* watch fewer files when running local tests
* changes unminified compile size from `3.7MB` to `3.2MB` (`1.6MB` minified)


mheiber:
@nelsonpecora why the switch to date-fns? Is it smaller?

nelsonpecora:
Yep, it's smaller _and_ more modular (you can require/import individual methods, similar to lodash)

nelsonpecora:
(also the API is really close to moment, and it's more functional)

PR #811, opened 2017-04-17
Expose store

mheiber:
- Expose Vuex store on `kiln` object

Expose the Vuex store so plugins and components can subscribe
to changes.

In particular, I think this may be needed for clay-space-edit.

Definitely open to suggestions if there is a better way of getting access to the store!

### DIFF
--- a/edit.js
+++ b/edit.js
@@ -65,6 +65,7 @@ Vue.directive('h-scroll', hScrollDirective());
 window.kiln = window.kiln || {};
 // .plugins, .behaviors, .validators, and .panes objects should already exist
 window.kiln.utils = utilsAPI;
+window.kiln.store = store;
 
 // kick off loading when DOM is ready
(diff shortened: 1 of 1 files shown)


mheiber:
@jonwinton ^

nelsonpecora:
> Definitely open to suggestions if there is a better way of getting access to the store!

There is! In kiln v1, if you add your plugin to `window.kiln.plugins`, it'll be passed the store in its initialization function (which will allow it to subscribe to mutations, grab state, and dispatch actions)

https://github.com/nymag/sites/blob/vue/global/kiln-js/plugins/word-count.js#L97

nelsonpecora:
(the thinking behind a discrete plugin api, rather than a global store, is that it's very easy to accidentally mutate the state and make Vue angry at you. Because of this, we're passing in only what we need for plugins (store), validators (store.state), etc)

mheiber:
ooh, much better. Thanks @nelsonpecora

PR #812, opened 2017-04-18
Fix clientSave to save new object returned by model

cperryk:
Currently, `clientSave` is returning the object that is passed into a model.save fnc instead of whatever object that fnc returns/resolves. As a result, a model save fnc that returns/resolves a new object instead of simply mutating the existing object will not work as expected. This PR fixes this bug. ([Discovered in work related to this Trello](https://trello.com/c/nEvvCuZN/105-imgur))


nelsonpecora:
!!! 👍 👍 👍

PR #818, opened 2017-04-19
expose window.kiln.services.getAvailableComponents

mheiber:
expose window.kiln.services.getAvailableComponents
that kind of thing is needed in clay-space-edit and could be useful for other plugins in the future.

Also want to note that this would be the first service in Kiln 1.0, and that a Vuex getter could be an OK alternative to using a service for this.


mheiber:
@nelsonpecora ^ per Slack conversation, thanks for taking a look!

nelsonpecora:
New Kiln has a different API strategy (instead of a 'services' thing, we're just exporting the specific utils we want in `utils/api`), so I added available components to the api in commit 6642cc6

You can access it via `kiln.utils.getAvailableComponents`

Issue #813, opened 2017-04-19
conditional-required


nelsonpecora:
* field B is required based on some logic with field A
* arguments could be `field`, `comparator` (defaults to "when field A is empty"), `value` (optional, if using some other comparitor)

Issue #815, opened 2017-04-19
reveal behavior


nelsonpecora:
* args: `field` to check, `value` to check against, `comparator` (optional)
* by default (no comparator), it's strict equality check
* special comparators: `isEmpty`, `regex`
* note: might want to share comparator logic with https://github.com/nymag/clay-kiln/issues/813

PR #820, opened 2017-04-19
reveal and conditional-required

nelsonpecora:
Now that we have Vue, we can do some cool behaviors that were impossible before. Announcing `reveal` and `conditional-required`!

# Reveal

* conditionally hide or show a field based on another field
* uses a new `comparators` library to keep things consistent

# Conditional Required

* conditionally makes a field required or not, based on another field
* uses `comparators` library
* includes a built-in validator


nelsonpecora:
this fixes #813 and #815

PR #864, opened 2017-06-12
from slate: update kiln 3 search textfields on panels to have some pa…

meijioro:
Hello from Slate,
I was working in the new Kiln 3 interface and the CSS side of me thought that the search textfield in the panels didn't look right so I updated it.

group

### DIFF
--- a/panes/filterable-list.vue
+++ b/panes/filterable-list.vue
@@ -4,10 +4,11 @@
 
   .filterable-list {
     &-input {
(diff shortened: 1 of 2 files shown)


mheiber:
🎉

nelsonpecora:
Hmm! I kinda like this. @gloddy your thoughts?

gloddy:
Looks like a winner to me.

PR #872, opened 2017-07-07
refactored pubsub to use a promise-based solution

nelsonpecora:
* this fixes the issue where a component wants to subscribe to properties published by separate components, but not wait for them all to publish before saving
* pubsub now has 100% test coverage
* the walls of the scrum room are covered in diagrams

img_0482

* also consolidated the `CURRENTLY_SAVING` stuff in the save action

### DIFF
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -26,6 +26,20 @@ function logSaveError(uri, e, store) {
 }
 
 /**
+ * re-render (reverting) a component and stop the saving promise chain
+ * @param  {string} uri
+ * @param  {object} data
+ * @param {string} html
+ * @param  {boolean} [snapshot]
+ * @param  {array} paths
+ * @return {Promise}
+ */
+function revertReject({ uri, data, html, snapshot, paths }) {
+  return store.dispatch('render', { uri, data, html, snapshot, paths })
+    .then(() => Promise.reject(new Error(`Save failed, reverting ${uri}`))); // don't continue saving
+}
+
+/**
  * save data client-side and queue up api call for the server
  * note: this uses the components' model.js and handlebars template
  * @param  {string} uri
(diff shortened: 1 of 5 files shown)


lib/component-data/pubsub.test.js
```diff
@@ -1,52 +1,62 @@
-import _ from 'lodash';
-import * as components from '../core-data/components';
 import * as lib from './pubsub';
 
-const schemas = {
-    noPub: { foo: 'a', bar: { baz: 'qux' }},
+const data = { foo: 'harder', bar: 'better', baz: 'faster', qux: 'stronger' },
```
elgreg: Listening to Daft Punk while writing this makes a lot of sense.

lib/component-data/pubsub.js
```diff
@@ -1,221 +1,159 @@
...
 import { pubProp, subProp, componentRoute, getComponentName } from '../utils/references';
 import { ADD_SCHEMA } from './mutationTypes';
 import { PRELOAD_SUCCESS } from '../preloader/mutationTypes';
-import { getSchema } from '../core-data/components';
-
-// Ask veit ek standa, heitir Yggdrasill
-// hár baðmr, ausinn hvíta auri;
-// þaðan koma döggvar þærs í dala falla;
-// stendr æ yfir grœnn Urðar brunni.
-export const yggdrasil = create();
```
mheiber: 😢

mheiber: Goodbye yggdrasil, thank you for making me smile

lib/component-data/actions.js
```diff
@@ -37,15 +38,12 @@ function logSaveError(uri, e, store) {
...
-  return modelSave(uri, data, {eventID, snapshot})
+  return modelSave(uri, data)
     .catch((e) => {
       // if model.js errors (or times out), re-render component with old data
       // and show the end user an error message
       logSaveError(uri, e, store);
-      return Promise.reject(store.dispatch('render', { uri, data: oldData, snapshot, paths }).then(() => {
-        store.commit(CURRENTLY_SAVING, false);
-      })); // don't continue saving
+      return Promise.reject(store.dispatch('render', { uri, data: oldData, snapshot, paths })); // don't continue saving
```
mheiber: Here `reject` is being called with the result of a dispatch, which doesn't seem like the same thing as a *reason*, which is the expected argument to `reject`: http://www.ecma-international.org/ecma-262/6.0/#sec-promise-reject-functions

It should be possible to render the old data and then reject the promise with a reason.

mheiber:
that whiteboard is beautiful

lib/component-data/actions.js
nelsonpecora: Hmm, good point

PR #894, opened 2017-08-23
esc / clicking out unfocuses BEFORE unselecting

nelsonpecora:
* back in old Kiln, clicking out (or hitting esc) would close the open form (unfocus) BEFORE unselecting the currently-selected component. this restores that functionality
* updates to selection when re-rendering: selected components that get re-rendered will still be selected afterwards. selected components that are removed when rendering will be unselected afterwards.
* better unselect handling when clicking around and selecting new components


nelsonpecora:
[Esc should unfocus forms, 2x Esc should unselect component](https://trello.com/c/akv7syPz/34-esc-should-unfocus-forms-2x-esc-should-unselect-component)

PR #896, opened 2017-08-24
use pages index to handle all page state

### LINKED ISSUE #895: Update page status checks
Before we had the built-in `pages` elastic index, we had to do a bunch of logic to check if a page was scheduled and/or published. Now that this exists, we should rely on it for all of the page status checks (and remove the old functionality).

nelsonpecora:
* completely removes old page state updating / preloading / logic
* uses `pages` index to determine pubbed/scheduled status in all places it's used
* should decrease server load and decrease initial kiln load time, since it won't do requests for the page@scheduled + page@published + article (if it exists) + published html (for the url check)
* fixes #895


mheiber:
screen shot 2017-08-25 at 9 54 27 am
😍

PR #898, opened 2017-08-28
remove server.js and non-handlebars support from kiln

### LINKED ISSUE #890: remove server-side component saving
* remove support for `server.js`
* remove support for non-handlebars templates

Both of these currently have a deprecation message, but should be removed in Kiln v4.0

nelsonpecora:
* BREAKING CHANGE for kiln v4.0.0 (closes #890)
* removes support for `serverSave` and `serverSaveAndRerender`
* removes assumption that components _need_ a `model.js` in order to do client-side optimistic saving + re-rendering (no more passthrough `model.js` files!)
* small bugfix for displaying page state correctly when viewing a published page
* small bugfix to optimize `kiln-view-public.js` rendering size (28k → 12k unminified)

### DIFF
--- a/lib/page-state/actions.js
+++ b/lib/page-state/actions.js
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import { postJSON } from '../core-data/api';
 import { UPDATE_PAGE_STATE } from './mutationTypes';
-import { searchRoute } from '../utils/references';
+import { searchRoute, replaceVersion } from '../utils/references';
 
 const postPageList = _.debounce(postJSON, 500); // only update the page list every 500ms
 
@@ -67,7 +67,7 @@ export function getListData(store, { uri, prefix }) {
       size: 1,
       query: {
         term: {
-          uri // match page uri
+          uri: replaceVersion(uri) // match page uri (removing @published or @scheduled)
         }
       }
     }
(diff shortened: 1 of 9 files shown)


lib/utils/shift-clay.js
```diff
@@ -1,4 +1,4 @@
-import { includes } from 'lodash';
+import _ from 'lodash';
```
byronhulcher: Maybe you should uhhhhhhh use `_.includes` for this later in the file? And also write tests? and maybe also test your code???????????????????

dmabuada: lol this is funny...how did this even..work?

lib/page-state/actions.js
```diff
@@ -67,7 +67,7 @@ export function getListData(store, { uri, prefix }) {
       size: 1,
       query: {
         term: {
-          uri // match page uri
+          uri: replaceVersion(uri) // match page uri (removing @published or @scheduled)
```
dmabuada: replace version or remove version? the pages index doesn't have any version appended its document keys

nelsonpecora: `replaceVersion` is the name of the function; if passed a single argument it'll remove any version that's applied (or just pass through the uri, if it doesn't have a version)

lib/utils/shift-clay.js
nelsonpecora: shift-clay.js is pretty simple logic, but would be non-trivial to test. thanks for catching that syntax error tho!

PR #899, opened 2017-08-29
add duplicate components

nelsonpecora:
* press ctrl or left command (on osx) to enable `state.ui.metaKey`, which we can use for a lot of additional functionality
* mini selector and expanded selector switch to "Duplicate Component" when meta key is pressed
* clicking "Duplicate Component" creates a new component of the same type as the currently selected, right after the selected component
* updated add components pane, add component (when only one type is allowed), and duplicate component to _select_ the newly added component afterwards

8be3bd90-2712-4f99-ba28-a84c58c6b36a-608-0000b7c49399f33e

51bb30a5-6bc3-44d8-9de3-17bf44df03fc-608-0000b7cf2605f7c2

### DIFF
--- a/edit.js
+++ b/edit.js
@@ -17,6 +17,7 @@ import hScrollDirective from './directives/horizontal-scroll';
 import utilsAPI from './lib/utils/api';
 import { hasClickedFocusableEl } from './lib/decorators/focus';
 import { hasClickedSelectableEl } from './lib/decorators/select';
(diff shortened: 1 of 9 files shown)


nelsonpecora:
[[key] + click on 'plus' button creates new component of same type as selected](https://trello.com/c/ojMvbuXR/35-key-click-on-plus-button-creates-new-component-of-same-type-as-selected)

PR #900, opened 2017-08-30
add people to current page

nelsonpecora:
* allows adding other people to the current page
* allows you to search through directory of people, to find the best person
* this does not notify the person added, but does put the page in their "My Pages" list

### DIFF
--- a/panes/add-person-to-page.vue
+++ b/panes/add-person-to-page.vue
@@ -1,0 +1,155 @@
+<style lang="sass">
+  @import '../styleguide/inputs';
+  @import '../styleguide/colors';
+  @import '../styleguide/buttons';
+  @import '../styleguide/typography';
+
+  // add-person-to-page is very similar to directory, but with fewer features
+  .add-person-to-page-pane {
+    height: calc(100% - 51px);
+
+    &-input {
+      padding: 4px;
+      &-field {
+        @include input();
+
+        padding: 10px 14px;
+      }
+    }
+
+    &-readout {
+      overflow-y: scroll;
+      overflow-x: hidden;
+      // height is pane height minus search box and bottom button area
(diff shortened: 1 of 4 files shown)


panes/add-person-to-page.vue
```diff
@@ -0,0 +1,155 @@
...
+          this.users = users;
+        });
+      },
+      onSearchKeyup: _.debounce(function () {
+        this.fetchUsers();
+      }, 300),
+      addPersonToPage(user) {
+        this.$store.dispatch('startProgress', 'save');
+        return this.$store.dispatch('updatePageList', { user }).then(() => {
+          this.$store.dispatch('finishProgress', 'save');
```
mheiber: I'd recommend adding `finishProgress` and closing the pane and such to the `catch` for the promise as well, unless the error handling is already happening elsewhere

nelsonpecora: Good point, I need to catch that if it fails for some reason

panes/add-person-to-page.vue
```diff
@@ -0,0 +1,155 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+  import _ from 'lodash';
+  import { postJSON } from '../lib/core-data/api';
+  import { searchRoute } from '../lib/utils/references';
+
+  function buildUserQuery(query) {
```
mheiber: nit:
buildUserQuery isn't a UI thing—by pulling it out of the Vue file it would be easier to test

mheiber:
code LGTM modulo 2 nits. Want me to run it and test in a browser?

panes/add-person-to-page.vue
nelsonpecora: Eh, the logic for it is super simple. There's not much to test 😅

PR #901, opened 2017-08-30
Adding, Removing, and Editing Page Templates

nelsonpecora:
* admins will see special "New Page" tab with additional options
* ability to "add current page to list" (with a specific name)
* ability to go directly to template (to edit it)
* ability to remove template from list (with confirmation)
* when editing a template directly, a (dismissible) status message appears
* when pressing ctrl / command, "add current page to list" becomes "create new page from current page"
* refactored list handling in kiln, adding methods to update lists easily (and putting api stuff in the api service)

screen shot 2017-08-30 at 3 47 55 pm


nelsonpecora:
I'm not a huge fan of the status message color, but it looks like we're using the same orangish color for both warnings (here, and in the health tab) and scheduling. @gloddy thoughts?

nelsonpecora:
[Add / remove / edit page templates](https://trello.com/c/aMWPbsPo/36-add-remove-edit-page-templates)

PR #902, opened 2017-09-01
complex list

nelsonpecora:
* MINOR new behavior, `complex-list`
* allows array of arbitrary fields
* MINOR export validation helpers in client-side api (so custom validators can use them)

## What is a Complex List?
* array of objects
* each object has properties defined by the schema
* yes, complex-lists can include complex-lists

## How does this work with behaviors that look at other fields' data?

* `reveal`, `magic-button`, and `conditional-required` now all use the same consolidated code, which looks at the _current list item_, then field in the _current form_, then field in the _current component_ to grab data.
* placeholders are displayed if there are no list items _or if all list items' fields are empty_
* placeholders, validators, and comparitors now use consolidated code to determine if a field is empty

## What does this look like?

923cb8b9-7e8f-417e-b6fd-c24101e6cc94-608-00014666afca3bdb

### DIFF
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -157,8 +157,8 @@ Append a magic button to an input.
 
 ## Arguments
 
-* **field** _(optional)_ a field to grab the value from
-* **component** _(optional)_ a name of a component to grab the component ref from
+* **field** _(optional)_ a field to grab the value from (in the current complex list, form, or component)
+* **component** _(optional)_ a name of a component to grab the component ref/uri from
 * **transform** _(optional)_ a transform to apply to the grabbed value
 * **transformArg** _(optional)_ an argument to pass through to the transform
 * **store** _(optional)_ to grab data from the client-side store
--- a/behaviors/README.md
+++ b/behaviors/README.md
@@ -80,7 +80,7 @@ A component which represents a single item in the `complex-list` component. Func
 
 # complex-list
 
-An array of objects with arbitrary properties. Each property may have any behaviors a field is allowed to have, including custom behaviors. Complex-list is similar to Angular's _transcluded directives_ or Advanced Custom Fields' _repeater field_, in that each item in the list is treated like a separate field. Items may also have `_label`, but may not have `_display` or `_placeholder`.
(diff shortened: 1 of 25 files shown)


lib/component-data/model.js
```diff
@@ -14,7 +14,7 @@ const MODEL_SAVE_TIMEOUT = 8000, // timeout for model.js pre-save hook
 export function save(uri, data) {
   const model = getModel(uri);
 
-  if (_.isFunction(model.save)) {
+  if (_.isObject(model) && _.isFunction(model.save)) {
```
amycheng: I'm curious about why you changed this?

behaviors/README.md
```diff
@@ -74,6 +74,26 @@ The mode of the editor sets syntax highlighting, linting, and other functionalit
 
 _Note:_ We will add more supported modes as we find use cases for them. See [this link](http://codemirror.net/mode/) for the full list of modes supported in codemirror.
 
+# complex-list-item
+
+A component which represents a single item in the `complex-list` component. Functionality is derived from its parent. Behaves similar to a field, as it transcludes other behaviors via the same "slot" mechanism that field uses.
+
+# complex-list
+
+An array of objects with arbitrary properties. Each property may have any behaviors a field is allowed to have, including custom behaviors. Complex-list is similar to Angular's _transcluded directives_ or Advanced Custom Fields' _repeater field_, in that each item in the list is treated like a separate field. Items may also have `_label`, but may not have `_display` or `_placeholder`.
```
amycheng: I think it would be helpful for there to be links to the Angular docs about transclusion or ACF's docs about the repeater field

lib/component-data/model.js
nelsonpecora: `model` will be undefined if a component doesn't have a `model.js`, but I hadn't caught it before (since beforehand we'd assume it had a server.js, or otherwise required server-side checks for logic)

behaviors/README.md
nelsonpecora: Good idea

PR #912, opened 2017-09-13
add create method to public api

amycheng:
... this will allow validators, plugins, behaviors and panes to create components without triggering a state change in `kiln`

While the `create` method is exposed via the `kiln` api, the logic for figuring out where the component does is NOT exposed and it is up to the developer to write this logic.  ie. which `component-list` does this component live?

### DIFF
--- a/lib/utils/api.js
+++ b/lib/utils/api.js
@@ -1,5 +1,5 @@
 import * as caret from './caret';
-import * as create from './component-data/create';
+import * as create from '../component-data/create';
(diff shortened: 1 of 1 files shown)


nelsonpecora:
Yep, that makes sense! This would allow plugins to do create components, put them into a created parent component, and add all of those to another component via the `addComponents` action.

nelsonpecora:
LGTM :shipit:

PR #915, opened 2017-09-13
Save empty array when adding cmpt list if prop is not set

cperryk:
Currently, if a component list property is not set in a cmpt instance, attempts to add a child to the list fail without a meaningful error message. (`Uncaught (in promise) TypeError: Cannot read property 'getAttribute' of undefined`)

This change modifies the component-list decorator to add an empty array to the cmpt-list prop if it has no current value.

### DIFF
--- a/lib/decorators/component-list.js
+++ b/lib/decorators/component-list.js
@@ -163,6 +163,7 @@ function addComponentList(el, componentList) {
  * decorate component list if specified in the schema
  * @param  {string} uri
  * @param  {Element} el with data-editable
+ * @returns {Promise|null}
  */
 export default function handler(uri, el) {
   const path = el.getAttribute(editAttr),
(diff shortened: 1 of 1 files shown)


lib/decorators/component-list.js
```diff
@@ -163,12 +163,16 @@ function addComponentList(el, componentList) {
...
+ * @returns {Promise|null}
  */
 export default function handler(uri, el) {
   const path = el.getAttribute(editAttr),
     schema = getSchema(uri, path); // note: component lists are ALWAYS a single field, so we don't need co call groups.get
 
   if (schema && schema[componentListProp]) {
     addComponentList(el, schema[componentListProp]);
+    if (getData(uri, path) === undefined) {
+      return store.dispatch('saveComponent', { uri, data: { [path]: [] }});
```
nelsonpecora: Hmm, the stuff that calls this decorator doesn't expect it to return a promise, but this is a valid use case. Let me take a look today / tomorrow morning and update this to cover all edge cases (e.g. if the list is null, or some other non-array value).

nelsonpecora:
I think the real answer to this is for the add component code to handle this situation, rather than triggering a re-save of the component

PR #916, opened 2017-09-14
github issue template and separate contributing doc

amycheng:
... to make it more welcoming and easier for folks to contribute to the project


amycheng:
I added a CoC (just copied JSConf's) but realized that there's no easy way for a contributor to report harassment or who they can go to.

PR #917, opened 2017-09-14
compare multiple fields

nelsonpecora:
* compare more than two fields in `ifEmpty` statements (when creating placeholders for groups)
* based on an idea / reference implementation that @byronhulcher did (thanks Byron!)
* allows comparing more than two fields _only if using identical comparators_ (e.g. `foo and bar and baz`, not `foo and bar or baz`)
* `xor` can logically only compare two booleans

### DIFF
--- a/lib/decorators/placeholders.js
+++ b/lib/decorators/placeholders.js
@@ -8,22 +8,49 @@ import placeholderTpl from './placeholder.vue';
 const Placeholder = Vue.component('placeholder', placeholderTpl);
 
 /**
- * compare two fields to see if they're empty
- * comparators are case insensitive, so 'AND', 'and', 'OR', 'or', etc work
- * @param {array} tokens must have 3 items: first field, comparator, second field
- * @param {object} fields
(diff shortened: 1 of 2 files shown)


lib/decorators/placeholders.js
```diff
@@ -8,22 +8,49 @@ import placeholderTpl from './placeholder.vue';
...
+ * @param {object} fields object with data for all fields
+ * @param {string} statement from ifEmpty, used for giving more information to error messages
  * @returns {boolean}
  */
-export function compareTwoFields(tokens, fields) {
-  const isFirstEmpty = isEmpty(fields[_.head(tokens)]),
-    comparator = tokens[1].toLowerCase(), // comparator is case-insensitive
-    isSecondEmpty = isEmpty(fields[_.last(tokens)]);
+export function compareFields(tokens, fields, statement) {
+  const fieldNames = _.filter(tokens, (token, index) => index % 2 === 0), // first, third, etc item (in 0-indexed array)
```
amycheng: love modulo

PR #927, opened 2017-10-19
 Make TK validator accept URIs

mheiber:
Make tk validation not fail on URIs.
Passes both URIs like slate.com and URIs
like localhost/components/foo/instances/00tk00

Fixes validation fail on component data like this:

screen shot 2017-10-19 at 3 48 17 pm

But also fixes TK warn for bare URLs in articles. Example: "the domain bantk.com is available."

### DIFF
--- a/lib/validators/built-in/tk.js
+++ b/lib/validators/built-in/tk.js
@@ -8,10 +8,14 @@ export const label = 'TKs',
   description = 'There are TKs in your article. Make sure they\'re intentional',
   type = 'warning';
 
+function isUri(str) {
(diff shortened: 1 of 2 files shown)


mheiber:
cc @elgreg and @felkerch

mheiber:
@nelsonpecora just wanted to check in again about this PR. We saw spurious warnings confuse both a PM and a designer in a dev environment, so this would fix a real problem.

nelsonpecora:
Sorry it took a bit!

PR #928, opened 2017-10-23
inject the Vuex store on mount so `Selector`'s children have access to it

amycheng:
for [reference](https://forum.vuejs.org/t/this-store-undefined-in-manually-mounted-vue-component/8756)

also, going to hold off on merging this until I've QA'd extensively on my local machine


amycheng:
k, tested it on my end. merging

PR #999, opened 2017-11-14
third-party custom buttons have access to Kiln utilities and Keen UI components

amycheng:
**Problem**:
the selector template was imported BEFORE `window.kiln.selectorButtons` was populated by any plugins that were loaded AFTER Kiln

**Solution**:
Vue templates are basically JSON objects, so add the custom buttons to the selector template `components` property during a function that is run after `DOM.contentloaded`


amycheng:
@nelsonpecora so the problem is not so much that the decorators are instantiated too soon but that `select.js` is imported (see [here](https://github.com/clay/clay-kiln/blob/master/edit.js#L16)) before the `DOMContentLoaded`.

When `select.js` is loaded at the top of `edit.js`, it immediately imports the selector template, so even if the decorators are loaded at the correct time, the selector template will still have the incorrect custom buttons.

In this latest commit, I'm using `$mount` to add the custom buttons, so it's less hacky.

nelsonpecora:
Ideally I'd like the whole `decorators` file to be instantiated at DOMContentLoaded, since there are potentially other things that might have to wait until then. (Hence the suggestion I made to do that `require` in `edit.js` instead)

amycheng:
I'm hesitant about using requires because we'll be using two methods for
loading modules (import and require). I get that webpack just converts the
imports into requires anyway but I'd prefer it be consistent in the
pre-compiled code.

But if you're fine with it, I'll work on it.

lib/decorators/select.js
```diff
@@ -81,7 +79,9 @@ export function getParentField(child, parent) {
  * @param {element} el  root element of component
  */
 export function addSelector(uri, el) {
-  let selector = new Selector({ store, componentEl: el }).$mount();
+  const selectorTpl = require('./selector.vue'),
```
amycheng: because dynamic imports don't exist T_T

amycheng:
@nelsonpecora used `require` to bring in the decorators and the Selector template when the DOM is ready. Basically just re-arranged some things

PR #1022, opened 2017-11-29
use colorful default avatars with initials

nelsonpecora:
Instead of the bland, grey user icon (or google's default avatar), use colorful circles with the user's initials if they don't have an avatar image

screen shot 2017-11-29 at 4 58 31 pm

screen shot 2017-11-29 at 3 05 56 pm

screen shot 2017-11-29 at 5 00 39 pm

### DIFF
--- a/lib/nav/collaborator.vue
+++ b/lib/nav/collaborator.vue
@@ -10,7 +10,7 @@
 
 <template>
   <div class="collaborator" @click.stop="selectCollaborator">
-    <avatar class="collaborator-image" :url="user.imageUrl" size="small" ref="avatar"></avatar>
+    <avatar class="collaborator-image" :url="user.imageUrl" size="small" :name="user.name || user.username" ref="avatar"></avatar>
     <ui-tooltip trigger="avatar" position="bottom left">{{ user.name || user.username }}</ui-tooltip>
   </div>
 </template>
--- a/lib/utils/avatar.vue
+++ b/lib/utils/avatar.vue
@@ -1,42 +1,51 @@
 <style lang="sass">
-  @import '../../styleguide/colors';
-
-  .avatar {
-    border-radius: 50%;
-    flex: 0 0 40px;
-    height: 40px;
-    width: 40px;
-
-    &.avatar-small {
-      flex: 0 0 24px;
-      height: 24px;
-      width: 24px;
-    }
-
-    &.avatar-default {
-      align-items: center;
-      background-color: $md-grey-500;
-      color: $pure-white;
-      display: flex;
-      justify-content: center;
-    }
+  .kiln-avatar {
+    flex: 0 0 auto;
   }
 </style>
 
 <template>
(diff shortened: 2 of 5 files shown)


lib/nav/collaborator.vue
```diff
@@ -10,7 +10,7 @@
 
 <template>
   <div class="collaborator" @click.stop="selectCollaborator">
-    <avatar class="collaborator-image" :url="user.imageUrl" size="small" ref="avatar"></avatar>
+    <avatar class="collaborator-image" :url="user.imageUrl" size="small" :name="user.name || user.username" ref="avatar"></avatar>
```
amycheng: why is the username in consistent?

lib/utils/avatar.vue
```diff
@@ -1,42 +1,51 @@
...
-  <img v-if="url" class="avatar" :class="'avatar-' + size" :src="url" />
-  <div v-else class="avatar avatar-default" :class="'avatar-' + size">
-    <ui-icon icon="person"></ui-icon>
-  </div>
+  <avatar class="kiln-avatar" :username="username" :src="imageURL" :size="pixelSize" :rounded="true"></avatar>
 </template>
 
 <script>
-  import UiIcon from 'keen/UiIcon';
+  import _ from 'lodash';
```
amycheng: don't think we need to load all of lodash

lib/utils/person.vue
```diff
@@ -62,7 +62,7 @@
 
 <template>
   <div class="person-item" :class="{ 'has-primary-action': hasPrimaryAction }">
-    <avatar class="person-image" :url="image" @click.stop="onClick"></avatar>
+    <avatar class="person-image" :url="image" :name="name || subtitle" @click.stop="onClick"></avatar>
```
amycheng: where is this `subtitle` property coming from?

lib/nav/collaborator.vue
nelsonpecora: `username` is what the 3rd party avatar component wants, but it means something different to use because we're _usually_ passing in the person's display name rather than their email

lib/utils/avatar.vue
nelsonpecora: kiln intelligently compiles only the parts of lodash we use

https://github.com/clay/clay-kiln/blob/master/webpack.config.js#L15

https://github.com/clay/clay-kiln/blob/master/.babelrc#L16

lib/utils/person.vue
nelsonpecora: `subtitle` is the secondary text that's passed into `person` components: https://github.com/clay/clay-kiln/blob/master/lib/utils/person.vue#L68

amycheng:
TIL!

lib/utils/avatar.vue
nelsonpecora: Yeah, the combination of packages is actually super cool and means we don't have to worry about which bits of lodash we're using

PR #1041, opened 2017-12-04
client-side layout re-rendering

### LINKED ISSUE #958: Update Page and Layout Data
Now that all layouts re-render client-side, we need to update the saving/rendering loop to re-enable support for changing component lists in the layout (including changing page data).

## Triggers

* component list dragdrop
* head components drawer dragdrop
* invisible components drawer dragdrop
* component list add / remove
* head components drawer add / remove
* invisible components drawer add / remove

## Saving Data

* saving layout data
* saving page data

## Rendering

* rendering head lists
* rendering body lists (including invisible lists)

nelsonpecora:
fixes #958

* intelligently re-render head components
* intelligently re-render head component lists in the layout
* intelligently re-render body component lists in the layout
* re-render head component lists in the page (from the server)
* re-render body component lists in the page (from the server)

### DIFF
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -445,6 +445,7 @@ export function openAddComponent(store, { currentURI, parentURI, path, pos }) {
       path: path,
       components: [{ name: _.head(available) }]
     }).then((newEl) => {
+      // select component if it isn't a layout or head component
       if (newEl) {
         store.dispatch('select', newEl);
       }
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -43,13 +43,12 @@ function logSaveError(uri, e, data, eventID, snapshot, store) { // eslint-disabl
  * re-render (reverting) a component and stop the saving promise chain
  * @param  {string} uri
  * @param  {object} data
- * @param {string} html
  * @param  {boolean} [snapshot]
  * @param  {array} paths
  * @return {Promise}
  */
-function revertReject({ uri, data, html, snapshot, paths }) {
-  return store.dispatch('render', { uri, data, html, snapshot, paths })
+function revertReject({ uri, data, snapshot, paths }) {
+  return store.dispatch('render', { uri, data, snapshot, paths })
     .then(() => Promise.reject(new Error(`Save failed, reverting ${uri}`))); // don't continue saving
 }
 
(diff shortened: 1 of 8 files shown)


lib/component-data/actions.js
```diff
@@ -436,7 +444,11 @@ export function openAddComponent(store, { currentURI, parentURI, path, pos }) {
       parentURI,
       path: path,
       components: [{ name: _.head(available) }]
-    }).then((newEl) => store.dispatch('select', newEl)));
+    }).then((newEl) => {
```
amycheng: when would `newEl` not be returned?

lib/component-data/reactive-render.js
```diff
@@ -128,25 +127,30 @@ function updateRootDOM(uri, newEl, paths) {
 export function updateDOM(uri, newEl, paths) {
   const isElement = newEl.nodeType === newEl.ELEMENT_NODE,
     isFragment = newEl.nodeType === newEl.DOCUMENT_FRAGMENT_NODE,
-    isDocument = newEl.nodeType === newEl.DOCUMENT_NODE;
+    isDocument = newEl.nodeType === newEl.DOCUMENT_NODE,
+    firstChild = isFragment && newEl.firstElementChild,
+    isFullPageFragment = firstChild && firstChild.tagName === 'META' && firstChild.getAttribute('charset') === 'utf-8';
```
amycheng: Isn't this making the assumption that a `layout`-type component will always start out like this? :
```
<!DOCTYPE html>
<html lang="en" data-uri="{{ default _ref _self }}" data-layout-uri="{{ default _ref _layoutRef }}">
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
...
```
but I can imagine someone dumping a `<style>` or `<script>` tag before that `<meta>` tag for a special kind of layout.

lib/component-data/actions.js
nelsonpecora: updating head components, or updating the layout

lib/component-data/reactive-render.js
nelsonpecora: Hmm, point. Let me see if there's another way to do this.

lib/component-data/actions.js
amycheng: could you add a code comment here? maybe something like "automatically select the new component if its not a layout or head component"

nelsonpecora: added

PR #1045, opened 2017-12-05
allow multiple attached buttons

### LINKED ISSUE #1000: Allow multiple attached buttons
Should we allow multiple attached buttons in inputs? It would allow things like this use case:

> FeedImgUrl allows you to fetch the first image component's url (via magic-button) or browse for one using the mediaplay picker

They might look a little weird if we allowed _n-number_ of buttons. Perhaps we could limit it to 2 or 3?

nelsonpecora:
* fixes #1000
* allows `n`-number of attached buttons (in practice people use one or two)
* will allow adding a `cancel` button to datepicker if they don't decide on a built-in solution for clearing dates

### DIFF
--- a/inputs/attached-button.vue
+++ b/inputs/attached-button.vue
@@ -1,0 +1,55 @@
+<template>
+  <div class="kiln-attached-buttons ui-button-group" v-if="args.attachedButton">
+    <component v-for="button in buttons" :key="button.name" slot="icon" :is="button.name" :name="name" :data="data" :schema="schema" :args="button" @disable="onDisable" @enable="onEnable"></component>
+  </div>
+</template>
+
+<script>
+  import _ from 'lodash';
+  import logger from '../lib/utils/log';
+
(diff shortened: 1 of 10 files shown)


inputs/attached-button.vue
```diff
@@ -0,0 +1,55 @@
...
+    },
+    computed: {
+      buttons() {
+        const attachedButton = _.get(this, 'args.attachedButton');
+
+        let buttons;
+
+        // munge them into an array
+        if (!!attachedButton) {
+          buttons = _.isArray(attachedButton) ? attachedButton : [attachedButton];
```
amycheng: maybe add `log.warn` if there are more than 2 buttons?

nelsonpecora: Eh, it's not recommended in most cases (because it makes the input smaller), but there's nothing actually wrong with using more than two attached buttons

Issue #1047, opened 2017-12-06
Validating on _componentList fields


zonika:
I need to be able to add validation to component list fields in kiln. For example, if I have a componentList that is required and needs to be between 5 and 300 items long, it would be nice to just specify that in the schema the way we do with input fields, rather than writing a custom plugin for kiln (which I am doing now).

nelsonpecora:
I don't want to add another validator until we standardize the `helper` methods, since otherwise we're just re-implementing a bunch of code (that honestly needs more testing as-is)

PR #1060, opened 2017-12-12
standardized validation helpers

### LINKED ISSUE #1015: add standardized validator helpers
Right now a lot of the validator code is custom to each validator, but that introduces some bugs and edge cases (especially around complex-list). We should consolidate some of the functionality into helpers which can be used by both built-in and 3rd party validators:

* check each field for a value (used by required, conditional-required, tk, etc)
* make sure deleted components don't get checked
* make sure fields that aren't defined properties in the data are checked (using the schema)
* check props inside complex-list items (returning the item's index for nice error messages)

nelsonpecora:
* implements methods to standardize validation
* updates built-in validators with new methods
* all new methods have 100% test coverage
* adds deprecation warning to old methods
* code cleanup

New Stuff!

* adds `min` validation
* renames `soft-maxlength` to simply `max`
* allows `min` and `max` validation on component lists #1047 
* adds `pattern` validation

screen shot 2017-12-12 at 2 41 39 pm

fixes #1015 #1061 #1047 #1062

### DIFF
--- a/lib/validators/helpers.js
+++ b/lib/validators/helpers.js
@@ -290,7 +290,7 @@ export function isValid(field, componentData, type) { // eslint-disable-line
     switch (type) {
       case 'required': return !(validate.required === true && isFieldEmpty);
       case 'conditional-required': return !(_.isObject(validate.required) && shouldBeRequired(field, componentData) && isFieldEmpty);
-      case 'pattern': return !(!strValue || validate.pattern && !strValue.match(new RegExp(validate.pattern, 'ig')));
+      case 'pattern': return !(!isFieldEmpty && validate.pattern && !strValue.match(new RegExp(validate.pattern, 'ig')));
       case 'min': return !(validate.min && (length || fieldValue) < validate.min);
       case 'max': return !(validate.max && (length || fieldValue) > validate.max);
       default: throw new Error(`Unsupported validation type: ${type}`);
@@ -299,7 +299,7 @@ export function isValid(field, componentData, type) { // eslint-disable-line
     // validate required fields first, then the rest of the built-in validation rules
     return !(validate.required === true && isFieldEmpty || // regular required
       _.isObject(validate.required) && shouldBeRequired(field, componentData) && isFieldEmpty || // conditional-required
-      validate.pattern && !strValue.match(new RegExp(validate.pattern, 'ig')) || // pattern validation
+      !isFieldEmpty && validate.pattern && !strValue.match(new RegExp(validate.pattern, 'ig')) || // pattern validation
       validate.min && (length || fieldValue) < validate.min || // minimum length / value
       validate.max && (length || fieldValue) > validate.max); // maximum length / value
   }
--- a/lib/validators/helpers.js
+++ b/lib/validators/helpers.js
@@ -286,7 +286,9 @@ export function isValid(field, componentData, type) { // eslint-disable-line
     length = (strValue || fieldValue).length;
   }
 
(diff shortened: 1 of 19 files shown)


lib/validators/helpers.js
```diff
@@ -30,6 +221,92 @@ export function getPreviewText(text, index, length) {
   return previewText;
 }
 
+/**
+ * detrmine whether a conditional-required field should be required
```
amycheng: typo

lib/validators/helpers.js
```diff
@@ -30,6 +221,92 @@ export function getPreviewText(text, index, length) {
...
+ * note: exported for testing
+ * @param  {object} field         { type, name, path, schema, value }
+ * @param  {object} componentData
+ * @return {boolean}
+ */
+export function shouldBeRequired(field, componentData) {
```
amycheng: I am not sure what is meant by "conditional-required field should be required". Could you describe a use case?

lib/validators/helpers.js
nelsonpecora: womp womp

lib/validators/helpers.js
nelsonpecora: whether that field should be required, based on the config passed in https://claycms.gitbooks.io/kiln/editing-components.html#standard-input-arguments

lib/validators/helpers.js
```diff
@@ -30,6 +221,92 @@ export function getPreviewText(text, index, length) {
...
+    switch (type) {
+      case 'required': return !(validate.required === true && isFieldEmpty);
+      case 'conditional-required': return !(_.isObject(validate.required) && shouldBeRequired(field, componentData) && isFieldEmpty);
+      case 'pattern': return !(!isFieldEmpty && validate.pattern && !strValue.match(new RegExp(validate.pattern, 'ig')));
+      case 'min': return !(validate.min && (length || fieldValue) < validate.min);
+      case 'max': return !(validate.max && (length || fieldValue) > validate.max);
+      default: throw new Error(`Unsupported validation type: ${type}`);
+    }
+  } else {
+    // validate required fields first, then the rest of the built-in validation rules
```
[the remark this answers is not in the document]
nelsonpecora: This doesn't come directly from schemas, but rather it's exposed to 3rd party helpers via the api.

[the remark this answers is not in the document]
amycheng: could you add a code comment about that? if someone is quickly skimming this code, this block looks like a different way to write the previous code block

lib/validators/helpers.js
```diff
@@ -30,6 +221,92 @@ export function getPreviewText(text, index, length) {
...
+  }
+
+  if (type) {
+    switch (type) {
+      case 'required': return !(validate.required === true && isFieldEmpty);
+      case 'conditional-required': return !(_.isObject(validate.required) && shouldBeRequired(field, componentData) && isFieldEmpty);
+      case 'pattern': return !(!isFieldEmpty && validate.pattern && !strValue.match(new RegExp(validate.pattern, 'ig')));
+      case 'min': return !(validate.min && (length || fieldValue) < validate.min);
+      case 'max': return !(validate.max && (length || fieldValue) > validate.max);
+      default: throw new Error(`Unsupported validation type: ${type}`);
```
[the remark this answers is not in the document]
nelsonpecora: I don't want it to log and continue, because this is a dev error that should (hopefully) never be hit unless someone's

1. in the process of writing a validator, that
2. erroneously assumes there's some kind of built-in validation that doesn't exist

Issue #1007, opened 2017-11-15
Unintended author names and tags are saved and appear in the drop down.


nelsonpecora:
* Use list service.
* This only works for lists and not Elastic.
* Use allowRemove argument.
* Also add allowCreate argument to automatically add things to list in Kiln.
* Trim things that are automatically added.

Note: autocomplete might be getting a little more powerful (Update auto-complete behavior to query Elastic), so this deletion functionality should probably work with both lists and elastic indices. (What security / data-loss concerns do we have when deleting items from an elastic index?)

![delete](https://user-images.githubusercontent.com/447522/32862449-2496f0de-ca26-11e7-9482-c537f03828da.png)

nelsonpecora:
Per our discussion, I think we should combine the schema args into a single `modifyList` argument, which would:

1. add an `X` button on items, to remove them from the list
2. automatically add _new_ items to the list (via the store)

pfeytser:
###Acceptance Criteria

* When the autocomplete list appears, I can click the 'x' on a specific item, which will delete it. 
* The next time I type the same word, the item I deleted does not appear in autocomplete.

pfeytser:
Acceptance Criteria

- [ ] When autocomplete options appear, there is an X next to each item. 
- [ ] When I click the X, it deletes that option from the list so it does not appear in autocomplete anymore. 
- [ ] This works for authors and tags.

nelsonpecora:
@pfeytser note that this should only enable that UI affordance in fields where the option is enabled (article authors and article tags).

Also implementation note: there are other components using `simple-list` that would have to be updated if they wanted to enable this functionality (feature ledes, etc), and if kiln also _adds_ new items to the list then those components' `model.save()` methods would have to be updated to not save the items to the list a second time.

PR #1080, opened 2018-01-05
Allow users to remove items from a list so it doesn't show up as an `autocomplete` match

amycheng:
related to https://github.com/clay/clay-kiln/issues/1007


docs/inputs.md
```diff
@@ -1,4 +1,13 @@
 
+# `autocomplete`
+UI that reveals potential matches to an entered value, ie a value entered via `simple-list`.
+
+### Autocomplete Arguments
+* **list** - name of a list that Amphora knows about accessible via `/<site>/_lists/<listName>`.
+* **allowDestroy** - boolean that determines whether the user can delete items from the list (so it doesn't appear as a match in the autocomplete). THIS IS SET VIA THE PARENT OF AUTOCOMPLETE.
```
nelsonpecora: I don't really understand this line. Is this specified in a schema? If not, I'd remove the autocomplete docs and leave the `modifyList` information in the relevant inputs, so as to not confuse devs

docs/inputs.md
```diff
@@ -441,6 +450,7 @@ An array of strings (or objects with a `text` property, if you add the `property
 * **badge** - name of the icon (or a two-character string) that should be displayed in the simple list item when editing. Icon names can be anything from the [Material Design Icon Set](https://material.io/icons/), or you can use two initials
 * **allowRepeatedItems** - allow the same item more than once. defaults to false
 * **autocomplete** - object with autocomplete options. Currently this is just the key `list` where the value is the name of a list that Amphora knows about accessible via `/<site>/_lists/<listName>`. Example:
+* **modifyList** - allow the user modify the list that is passed into autocomplete. This will send a PUT to Amphora. set `allowDestory` to true, to let the user remove items from the list via the autocomplete UI
```
nelsonpecora: Per above, I wouldn't get into the internal details of `allowDestroy`. Rather, just discuss the exposed configuration option and what it does from a user perspective (enabling a UI affordance to remove items from the autocomplete list)

Also, will new items automatically be _added_ to the list if this is enabled? I think that's what "this will sent a PUT to amphora" means, but it's not super clear.

By the by, you should probably change the line above to reflect the new configuration options allowed.

inputs/autocomplete-item.vue
```diff
@@ -29,18 +39,30 @@
     v-bind:class="{ 'selected': isActive }"
     type="button"
     @click.stop.prevent="select(value)">
-    {{ value }}
+    <span class="item-value">{{value}}</span>
+    <span class="item-actions">
+      <ui-icon-button v-show="canDestroy" icon="clear" :tooltip="`Remove ${value} from list`" @click.stop="destroy(value)" color="default" size="small" type="secondary"></ui-icon-button>
```
nelsonpecora: This is a little nitpicky, but the convention we've been using is putting the event handlers at (or near) the end of attributes. 😅

inputs/autocomplete-item.vue
```diff
@@ -29,18 +39,30 @@
...
-    props: ['value', 'index', 'select', 'focusIndex'],
+    props: ['value', 'index', 'select', 'focusIndex', 'destroy', 'data'],
     data() {},
     computed: {
       isActive() {
         return this.index === this.focusIndex;
+      },
+      canDestroy() {
+        // only show the remove button if the schema allows it
+        return this.data.allowDestroy;
```
nelsonpecora: instead of passing in `data`, would it be clearer to pass in `allowDestroy` directly? I got a bit confused looking at `value` and `data` together as props.

inputs/autocomplete.vue
```diff
@@ -22,17 +33,21 @@
...
+        :destroy="removeFromList"
+        ></item>
     </li>
   </ol>
 </template>
 
 <script>
   import _ from 'lodash';
   import item from './autocomplete-item.vue';
+  import {removeListItem} from '../lib/lists/helpers';
```
nelsonpecora: Another nitpick: spaces around destructured imports 😅

inputs/autocomplete.vue
```diff
@@ -90,22 +105,37 @@
...
+      removeFromList(item) {
+        const listName  = this.args.list,
+          unselect = this.unselect;
 
-      if (items) {
-        promise = Promise.resolve(items);
-      } else {
-        promise = this.$store.dispatch('getList', listName).then(() => _.get(lists, `${listName}.items`));
-      }
+        unselect();
```
nelsonpecora: I'm not sure why this needs to be a variable. It doesn't look like it's reused anywhere?

inputs/simple-list.vue
```diff
@@ -152,6 +153,15 @@
       };
     },
     computed: {
+      autocompleteOpts() {
+        let opts = this.args.autocomplete;
+
+        if (_.has(this.args, 'modifyList') && _.has(this.args.modifyList, 'allowDestroy')) {
```
nelsonpecora: Wait, I inferred from the docs that `modifyList` was a boolean, but that doesn't seem the case? It's an object? What other properties would go in that object?

lib/lists/helpers.js
```diff
@@ -0,0 +1,58 @@
...
+
+function getFormat(items) {
+  const tester = _.head(items);
+
+  if (_.isString(tester)) {
+    return 'STRING';
+  } else if (_.has(tester,'text')) {
+    return 'OBJECT';
+  } else {
+    log.error('List cannot be modified because it is not a simple list.', { action: 'modifyList' });
```
nelsonpecora: Is this actually an error? Per the spec, lists can be arrays of _arbitrary data_. They must be strings or objects (with `text`) to be used by autocomplete (which also should be documented), but this file is a helper for all lists, not just lists that autocomplete cares about.

lib/lists/helpers.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+/**
+ * @param {array} items
+ * @param {string} testItem
```
nelsonpecora: This says `testItem` is a string, but the code below says it may be an object (but no other type?)

lib/lists/helpers.js
```diff
@@ -0,0 +1,58 @@
...
+ * @returns {integer}
+ */
+export function getItemIndex(items, testItem) {
+  // force tester to be a string
+  const tester = _.isString(testItem) ? testItem : testItem.text,
```
nelsonpecora: Should this call `getFormat` here, rather than below? (since that function handles the case where it's neither a string nor object)

inputs/autocomplete.vue
amycheng: Made it a variable for a previous use case that's not applicable anymore. changing!

inputs/simple-list.vue
amycheng: It's an object. I was going to dump `allowCreate` in there. Going to work on that feature in another PR.

lib/lists/helpers.js
amycheng: `testItem` can be either a string OR object, changing.

docs/inputs.md
```diff
@@ -1,4 +1,11 @@
 
+# `autocomplete`
+UI that reveals potential matches to an entered value, ie a value entered via `simple-list`.
+
+### Autocomplete Arguments
```
nelsonpecora: Per [my previous comment](https://github.com/clay/clay-kiln/pull/1080#discussion_r159968270), these are internal arguments to `autocomplete` that aren't exposed as an input for devs to use, so they shouldn't be documented here.

docs/inputs.md
```diff
@@ -441,6 +448,7 @@ An array of strings (or objects with a `text` property, if you add the `property
 * **badge** - name of the icon (or a two-character string) that should be displayed in the simple list item when editing. Icon names can be anything from the [Material Design Icon Set](https://material.io/icons/), or you can use two initials
 * **allowRepeatedItems** - allow the same item more than once. defaults to false
 * **autocomplete** - object with autocomplete options. Currently this is just the key `list` where the value is the name of a list that Amphora knows about accessible via `/<site>/_lists/<listName>`. Example:
+* **modifyList** - options that allow the user to modify lists via Kiln. Setting `allowRemove` to true, enables an X in the `autocomplete` that allows the user to remove that item from the autocomplete list.
```
nelsonpecora: Shouldn't this (or the related `allowRemove` argument itself) be specified under `autocomplete` (the line above)?

inputs/autocomplete-item.vue
```diff
@@ -29,18 +39,30 @@
     v-bind:class="{ 'selected': isActive }"
     type="button"
     @click.stop.prevent="select(value)">
-    {{ value }}
+    <span class="item-value">{{value}}</span>
+    <span class="item-actions">
+      <ui-icon-button v-show="canDestroy" icon="clear" :tooltip="`Remove ${value} from list`" color="default" size="small" type="secondary" @click.stop="destroy(value)"></ui-icon-button>
```
nelsonpecora: You're using _destroy_ and _remove_ interchangeably. It would reduce developer mental overhead if a single term (say, _remove_) was used everywhere.

inputs/autocomplete.vue
```diff
@@ -22,17 +31,21 @@
         :index="index"
         :focusIndex="activeIndex"
         :value="match"
-        :select="select"></item>
+        :select="select"
+        :allowRemove="args.allowRemove"
+        :destroy="removeFromList"
```
nelsonpecora: e.g. right here

inputs/simple-list.vue
```diff
@@ -152,6 +153,15 @@
       };
     },
     computed: {
+      autocompleteOpts() {
+        let opts = this.args.autocomplete;
+
+        if (_.has(this.args, 'modifyList') && _.has(this.args.modifyList, 'allowRemove')) {
+          _.set(opts, 'allowRemove', this.args.modifyList.allowRemove);
```
nelsonpecora: by having the args look like (something like):

```
autocomplete:
  list: some-list
  allowRemove: true
```

this whole part of the code (and the api in general) would be a lot simpler

lib/lists/helpers.js
```diff
@@ -0,0 +1,73 @@
...
+  const tester = _.head(items);
+
+  if (_.isString(tester) && _.isObject(testItem) ||
+      _.isObject(tester) && _.isString(testItem) && !testProperty) {
+    log.error('The item you are looking for does not have the same data structure as the items in the list.', {action: 'modifyList'});
+    return false;
+  }
+
+  if (_.isObject(tester) && _.isObject(testItem)) {
+    listItemKeys = _.keys(_.head(items)).sort();
```
nelsonpecora: You don't need to do `_.head(items)` here again, because you're using `tester`, right?

PR #1081, opened 2018-01-08
Scoped PubSub

### LINKED ISSUE #1075: Scoped PubSub
So imagine you have a component who wants to update it's children components with a value. Cool, pub/sub. 

But what if you have multiple instances of that component and you want to scope the property that is being pubbed to _only that component's children_.

Angular had an interesting concept on it's `$scope` object with `$emit` and `$broadcast` indicating "direction". I wonder how big of a lift something like this might be. The idea of scoping brings inherent complexity, but also could make components more flexible.

nelsonpecora:
* fixes #1075
* allows limiting scope to parents or children
* allows scoped pubsub in arrays
* refactored pubsub for code clarity
* updated docs

Single property:

```yaml
slideshowWidth:
  _publish:
    name: width
    scope: children
  _has:
    help: Width should propagate only to children
```

Multiple properties:

```yaml
title:
  _has: text
  _publish:
    - ogTitle
    - twitterTitle
    - name: socialTitle
      scope: children
    - name: articleTitle
      scope: parents
```

### DIFF
--- a/lib/component-data/pubsub.js
+++ b/lib/component-data/pubsub.js
@@ -20,6 +20,21 @@ let callStacks = {},
   // this allows us to determine which component(s) should update when properties are published
 
 /**
+ * convert publishable props to objects (or arrays of objects)
+ * @param  {string|object|array} prop
+ * @return {object|array}
+ */
+function convertPubObjects(prop) {
+  if (_.isArray(prop)) {
+    return _.map(prop, (part) => _.isString(part) ? { name: part } : part);
+  } else if (_.isString(prop)) {
+    return { name: prop }; // convert to object
+  } else {
+    return prop; // already an object
+  }
+}
+
+/**
  * register publishers and subscribers for a component
  * note: exported for testing
  * @param  {string} name   of component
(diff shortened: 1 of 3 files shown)


lib/component-data/pubsub.js
```diff
@@ -132,6 +148,68 @@ function shouldSubscribe(name, eventID) {
...
+ * @param  {string} scope
+ * @param  {string} publisherURI
+ * @param  {string} subscriberURI
+ * @return {boolean}
+ */
+function subscriberInScope(scope, publisherURI, subscriberURI) {
+  if (scope === 'global') {
+    // 'global' scope is the default
+    return true;
+  } else if (_.includes(['children', 'child'], scope)) {
```
amycheng: why is it checking for both 'children' and 'child', 'parents' and 'parent'?

nelsonpecora: While `children` and `parents` is the documented, acceptable option, I didn't want to penalize people who might mistakenly type `child` or `parent` (since they're very easy to confuse _and_ are frequently used to describe this kind of scoping in other systems / frameworks)

PR #1097, opened 2018-01-22
Update Kiln Page History With Edit Events

### LINKED ISSUE #1094: Update Kiln Page History
Per https://github.com/clay/amphora-search/issues/54 Kiln should update `/_pagelist` when page-specific components are saved. It should:

* Add an `edit` action if the previous action is not `edit`.
* Add the current user to the `users` array of the previous action if it _is_ edit and the current user is _not_ in `users`
* Do nothing if the previous action is `edit` and the current user _is_ in the `users` array

byronhulcher:
Fixes #1094

### DIFF
--- a/lib/page-state/actions.js
+++ b/lib/page-state/actions.js
@@ -7,7 +7,36 @@ import { searchRoute, replaceVersion } from '../utils/references';
  * @module page-state
  */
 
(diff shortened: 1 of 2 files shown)


nelsonpecora:
This shouldn't be merged until we can add the `peerDependency` for the version of `amphora-search` it requires

Issue #1088, opened 2018-01-12
Add `rows` arg to textarea input


nelsonpecora:
The `multi-line` text input (textarea) currently defaults to two rows tall, but this should be configurable in component schemas.

```yaml
input: text
type: multi-line
rows: 4
```

PR #1103, opened 2018-01-29
add rows argument to tex input

amycheng:
implements https://github.com/clay/clay-kiln/issues/1088


inputs/text.vue
```diff
@@ -94,6 +96,9 @@
       },
       isInvalid() {
         return !!this.errorMessage;
+      },
+      numOfRows() {
+        return _.has(this.args,'rows') ? this.args.rows : 2;
```
nelsonpecora: ah, it might be easier to do `_.get(this, 'args.rows', 2)`

amycheng: good point changing.

PR #1113, opened 2018-02-08
Add props to handle select default option

frederickcxa:
Adds support for handling the default option `None` in the `selects` UI component.

Added properties:

- `excludeDefaultOption`: allow users to exclude the automatically added default option `None`, defaults to `false`.
- `defaultOptionLabel`: allow users to rename the default option's label, only if `excludeDefaultOption` is `true`.

e.g:
```
input: select
    options:
      - red
      - yellow

Renders:
- None (null)
- Red (red)
- Yellow (Yellow)
```

```
input: select
    defaultOptionLabel: White
    options:
      - red
      - yellow

Renders:
- White (null)
- Red (red)
- Yellow (yellow)
```

```
input: select
    excludeDefaultOption: true
    defaultOptionLabel: White
    options:
      - red
      - yellow

Renders:
- Red (red)
- Yellow (yellow)
```

### DIFF
--- a/inputs/select.vue
+++ b/inputs/select.vue
@@ -7,6 +7,8 @@
 
   * **multiple** - allow multiple options to be selected. data will be an object with options as keys, similar to checkbox-group
   * **search** - allow users to type stuff in to filter options. Extremely useful for longer options lists
+  * **excludeDefaultOption** - allow users to exclude the default option `None`, defaults to `false`.
+  * **defaultOptionLabel** - allow users to rename the default option's label.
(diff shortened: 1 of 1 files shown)


frederickcxa:
Thanks, at first the need was to remove the `empty/None` option (not visible to the user), but at the end with making it required was enough. This also lets the developer change the `None` item label. But, after requirements changes, removing the `None` option is not that useful as it was.

nelsonpecora:
Roger that, closing.

Issue #1121, opened 2018-02-15
generate _variation field in settings


nelsonpecora:
* parse `_componentVariations` on page load
* for components with variations, generate a `Variations` field in that component's settings
* on save, fetch new variation css if user has switched a variation

amycheng:
we should also account for components that have variations but NO settings group defined in their schemas

PR #1125, opened 2018-03-05
Kiln fetches component variations

amycheng:
... and automatically generates a UI that allows the user to select the variation


amycheng:
related to https://github.com/clay/clay-kiln/issues/1121

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,104 @@
+import _ from 'lodash';
+import logger from '../utils/log';
+
+const log = logger(__filename);
+
+/*
+ * adds a 'componentVariation' field to the components schema
+ * @param {schema} object
+ * @param {componentVariations} array
```
cperryk: Not the correct way to document functions

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/*
+ * create a settings group if the component does not have one
+ * @param {variations} object
+ * @param {componentName} string
+ * @returns {boolean}
+ */
+function hasVariations(variations, componentName) {
+  return _.has(variations, componentName);
```
cperryk: Why use this function over `_.has` directly?

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/*
+ * create a settings group if the component does not have one
+ * @param {schema} object
+ * @returns {object}
+ */
+function addVariationSection(sections) {
+  const newSectionTitle = sections.fields.length === 0 ? 'General' : 'Component Variation';
+  let newSections = sections;
```
cperryk: No reason this should be a `let` over a `const`

lib/core-data/get-variations.js
amycheng: had a brain fart, fixed

lib/core-data/get-variations.js
amycheng: thought I was going to use it more often, but nope. fixed

lib/core-data/get-variations.js
amycheng: fixed

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * adds a 'componentVariation' field to the components schema
+ * @param {object} schema
+ * @param {array} componentVariations
+ * @returns {object}
+ **/
+function addField(schema, componentVariations) {
+  const variations = _.map(componentVariations, function (variant) {
+      return {
+        name: _.capitalize(variant.split('_')[1]),
```
reubenson: do we care to replace dash-separated names with space-separated?

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,94 @@
...
+function addVariationField(schema, componentName, variations) {
+  if (_.has(variations, componentName)) {
+    return addField(schema, variations[componentName]);
+  } else {
+    return schema;
+  }
+}
+
+/**
+ * create a settings group if the component does not have one
```
reubenson: fn description doesn't match action

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {object} sections
+ * @returns {object}
```
reubenson: sections is an array, not object?

lib/core-data/get-variations.js
nelsonpecora: The underscore separates the variation from the component name. They're pulled originally from the css filenames.

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,94 @@
...
+function createSettings(schema) {
+  let processedSchema = schema;
+
+  if (!_.has(processedSchema, '_groups')) {
+    _.set(processedSchema, '_groups', {
+      settings: {
+        fields: []
+      }
+    });
+    log.info('Kiln created a settings group to this schema because it was missing.', { action: 'addSettings'});
```
nelsonpecora: do we need to log this every time? Maybe it could be a `debug`, if anything

lib/core-data/get-variations.js
nelsonpecora: `sections` might not be the best name for this variable (because of `sections.sections`)

nelsonpecora: maybe `form` or `group` or something?

lib/preloader/actions.js
```diff
@@ -131,6 +131,7 @@ export default function preload({ commit, dispatch }) {
     // add preloaded data to the store
     commit(PRELOAD_SUCCESS, _.assign(tabulaRasa, {
       components: _.assign(components, composeLayoutData(layoutSchema, components, data)),
+      componentVariations: _.get(data, '_componentVariations'),
```
nelsonpecora: you should add `componentVariations` to the default (`tabulaRasa`) file and document them there

lib/core-data/get-variations.js
nelsonpecora: Or `parsed`, since that's the name of the same variable in the function that calls this one

amycheng: @reubenson nope, `sections` (now called `parsed`) is an object

lib/preloader/actions.js
amycheng: done

lib/core-data/get-variations.js
```diff
@@ -0,0 +1,93 @@
...
+
+  if (_.has(processedSchema, 'componentVariation')) {
+    return processedSchema;
+  } else {
+    return _.set(processedSchema, 'componentVariation', variationField);
+  }
+}
+
+/**
+ * update a scheme with a settings group if the component does not have one
```
reubenson: typo 'schema'

PR #1130, opened 2018-03-08
Drop support for non-underscored URIs

cperryk:
Stable versions of other Clay resources (e.g. amphora-search, clay-utils) no longer support non-underscored URIs. Let's have Kiln follow suit.

### DIFF
--- a/template.handlebars
+++ b/template.handlebars
@@ -11,8 +11,6 @@
     <script>
       // start scaffolding kiln stuff
       window.kiln = window.kiln || {};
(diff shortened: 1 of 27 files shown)


template.handlebars
```diff
@@ -11,8 +11,6 @@
     <script>
       // start scaffolding kiln stuff
       window.kiln = window.kiln || {};
-      // figure out the route prefix (e.g. /components, or /_components) based on the _ref of kiln itself
-      window.kiln.routePrefix = '{{ default _ref _self }}'.indexOf('/_components/') > -1 ? '_' : '';
```
reubenson: 🙌

PR #1133, opened 2018-03-12
Lists for ui select

elgreg:
Hi! This PR allows for supporting three new options in the UiSelect input:

1. `list` - similar to the `list` option used for autocomplete, this will look for the list in /lists/ endpoint of your site and create the select options based on that list
2. `keys` - passthrough option for Keen to specify keys for input objects, especially for use when you don't control the input shape, e.g. lists. Defaults to `{label: 'name', value: 'value'}`
3. `storeRawData` - normally only the `value` of each option is stored, but with this option you can store the entire input object

The use-case for all three of these together is when you have a list on /lists/ that looks like:

```
rubrics:
  - sectionSlug: human-interest
    slug: relationships
    text: Relationships
  - sectionSlug: human-interest
    slug: work
    text: Work
  - sectionSlug: human-interest
    slug: care-and-feeding
    text: Care and Feeding
...
```

UiSelect can now fetch that list and use keys that don't have specific names. It can also store more than just the value (e.g. the entire resulting object).

PR #1141, opened 2018-03-29
Lists for ui select (#1133) for Kiln 5

macgyver:
Hey @nelsonpecora ! cc @elgreg we'd like to cherry-pick this commit into the v5 release - not sure what your process is for patching older releases so let us know how you prefer to do it!

Don't have a ton of open-source/npm experience but this is my guess about the process:
1. in clay repo, check out and push up a new `v5` "release" branch based on the current latest v5 tag, i.e. `git co -b v5 v5.16.0; git push -u`
2. reset the base of this PR to that branch, review and merge
3. from the `v5` release branch, `npm version minor; npm publish --tag v5.17.0`


jonwinton:
Kiln 6.0 was breaking because the API contract with Elastic had to change. You can definitely add back in support for the non-underscored route, but the Elastic part was the main thing.

PR #1142, opened 2018-03-29
Fixes health check drawer not updating

### LINKED ISSUE #1127: Health Drawer not updating
## Bug
Validation doesn't track the removal of paragraphs. Only the insertion.

## Recreate
Open the drawer for Publish/Health & add a paragraph to an article. As expected a 'required' warning appears for empty to paragraph.
Either add content or delete the paragraph. The warning remains. The only ways to have it track the change is close/re-open the Publish tab or reload the page.

### Background
This was discovered while filtering out empty paragraphs when switching between Standard Display & Mini Display.

cperryk:
Fixes #1127 

The problem is caused by the Drawer component attempting to update application state in one of its own data computation functions, rendering it unable to reliably detect relevant changes to application state, e.g. editing a new component will _not_ trigger a re-computation of the Drawer, which means validation state will not be re-computed. Revalidation needs to be more directly tied to the actions that should trigger it. As far as I understand, we need to revalidate (a potentially expensive operation) only in these scenarios:

1. User opens publish or health drawer
2. User updates, creates, or removes a component while publish/health drawer is open

One possible solution would be to add logic to some actions to revalidate if the drawer condition is met. An alternative approach, which I've taken in this PR, is to create a plugin that listens for relevant mutations. I slightly prefer this approach because it isolates the validation logic instead of spreading it among various actions.

### DIFF
--- a/lib/validators/plugin.js
+++ b/lib/validators/plugin.js
@@ -10,12 +10,7 @@ const relevantMutations = [
     UPDATE_PAGE
   ],
   relevantDrawers = ['health', 'publish'],
-  validateDebounced = _.debounce(validate);
-
-function validate(store) {
-  store.dispatch('validate')
-      .then(() => store.commit('VALIDATE'));
-}
+  validate = _.debounce(store => store.dispatch('validate'), 250);
 
 export default (store) => {
   store.subscribe(({type}, state) => {
@@ -24,11 +19,8 @@ export default (store) => {
     if (!relevantDrawers.includes(_.get(state, 'ui.currentDrawer'))) return;
 
     // only revalidate on mutations we care about
-    if (relevantMutations.includes(type)) {
-      validateDebounced(store);
-    } else if (type !== 'VALIDATE') {
-      // commit vanity mutation so a plugin could conduct its own validations
-      store.commit('VALIDATE');
-    }
+    if (!relevantMutations.includes(type)) return;
+
+    validate(store);
   });
 };
--- a/lib/validators/plugin.js
+++ b/lib/validators/plugin.js
@@ -1,9 +1,7 @@
 import _ from 'lodash';
-import { UPDATE_PAGE_STATE } from '../page-state/mutationTypes';
 import { UPDATE_COMPONENT, REMOVE_COMPONENT, CREATE_COMPONENTS } from '../component-data/mutationTypes';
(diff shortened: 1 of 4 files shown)


amycheng:
I like this approach of putting the validation logic in a plugin.

While triggering validation when updating, creating, or removing components covers most use cases, what if a developer wants to trigger validations during other mutations?

For example this PR will not cover these cases:

- validating the scheduled publish date
- adding/removing scopes to the page (per this [issue](https://github.com/clay/amphora/issues/527))
- validating during undo/redo
- real-time word count or spellcheck

So do we want to:
- limit when validation happens (like in this PR) and encourage developers to submit PRs if they want their validators to listen to more mutations? (if we go this route, it should be documented)
- try to account for more mutations and add them to the plugin?
- develop architecture to let third-party validators add to the list of `relevantMutations` (more work though)

Also, can you also confirm that re-ordering components triggers a mutation? I am not 100% sure that it does (I think it might be a bug?).

cperryk:
@amycheng 

* Validating scheduled publish date wouldn't work anyway because failed validation blocks publishing. If there's a validation error, the scheduled publish time field becomes non-editable.
* I don't think we need to consider scopes right now considering they're just a proposal, but if we do implement scopes we can simply add to `relevantMutations`.
* Undo/redo works with this approach because they execute the same mutations that this checks for.
* I don't think real-time word count or spellcheck relate to this bug fix. As far as I understand we've never discussed validators firing on live edits. IMO it's a broader feature discussion.

A slightly alternative approach would be to revalidate on all state changes but add debouncing. This still leads to a bunch of scenarios in which we're doing unnecessary revalidation and may negatively affect performance. Just mousing over a component changes application state.

lib/validators/plugin.js
```diff
@@ -0,0 +1,26 @@
...
+import { OPEN_DRAWER } from '../drawers/mutationTypes';
+const relevantMutations = [
+    UPDATE_COMPONENT,
+    OPEN_DRAWER,
+    REMOVE_COMPONENT,
+    CREATE_COMPONENTS,
+    UPDATE_PAGE
+  ],
+  relevantDrawers = ['health', 'publish'],
+  validate = _.debounce(store => store.dispatch('validate'), 250);
```
amycheng: could you add a comment here on why there's debouncing? What you said in the PR comment is good.

Issue #1144, opened 2018-03-29
Allow deep field comparisons in group placeholders


zonika:
I have a complex list field that is bootstrapped with one item already in it. One of the item's props is set by the bootstrap and the other is an empty array. I'd like to be able to show a placeholder if the array in the first item of the complex-list is empty. Here is an field:

```yaml
byline:
  _label: Byline
  _has:
    input: complex-list
    props:
      -
        prop: names
        _label: Byline
        _has:
          input: simple-list
          autocomplete:
            list: authors
            allowRemove: true
            allowCreate: true
      -
        prop: prefix
        _label: Prefix
        _has:
          input: radio
          options:
            - 
              name: By
              value: by
            - 
              name: As told to
              value: as-told-to
```
this field is bootstrapped with one item, `{ names: [], prefix: 'by' }` so just using `ifEmpty: byline` won't trigger the placeholder to show up. Ideally, the field's placeholder would look something like this:

```yaml
    _placeholder:
      text: Byline
      height: 30px
      ifEmpty: 'byline[0].names'
```
That way we show the placeholder if the list of names in the first item of the complex-list is empty. I imagine the logic for this would use something similar to lodash's _.get(), so that if `byline[0]` didn't exist, isEmpty would still be true.

nelsonpecora:
Per Zoe, this is a good compromise and she's gonna take a crack at it. Assigning to her.

nelsonpecora:
released in 6.2.0

PR #1145, opened 2018-03-30
allow deep field comparisons in field and group placeholder logic

zonika:
[issue](https://github.com/clay/clay-kiln/issues/1144)


lib/decorators/placeholders.test.js
```diff
@@ -14,6 +14,29 @@ describe('placeholders', () => {
     sandbox.restore();
   });
 
+  describe('isFieldEmpty', () => {
+    const fn = lib.isFieldEmpty;
+
+    it('throws error if path depth is greater than 2', () => {
+      expect(() => fn({}, 'one.two.three')).to.throw(Error);
```
nelsonpecora: we should support n-depth nesting, since complex-lists may exists inside of other complex-lists

Issue #1160, opened 2018-04-05
BUG: unable to set default variation on a component


amycheng:
## Bug
If a user sets a variation on a component, but then wants it to go back to the default, they are unable to do so, because it is not an option in the select.

PR #1161, opened 2018-04-06
fixes a bug where the user cannot reset a component after a variation has been chosen

amycheng:
... also lets user assign a value to the 'None' option in select drop-downs
Closes https://github.com/clay/clay-kiln/issues/1160 and https://github.com/clay/clay-kiln/issues/1154

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -60,4 +60,4 @@ Schemas for components that are currently on the page are automatically parsed b
 
 ## Variations
 
-Components can have variations, which is defined by corresponding stylesheets in the site's styleguide folder. There is the default variation (`componet-name.css`)and variations are defined as `component-name_variation.css`. Kiln will automatically generate a select input with the variations and a 'Component Variation' tab in the Settings form. If the component doesn't have Settings, Kiln will generate this too.
+Components can have variations, which is defined by corresponding stylesheets in the site's styleguide folder. There is the default variation (`componet-name.css`) and variations are defined as `component-name_variation.css`. Kiln will automatically generate a select input with the variations and a 'Component Variation' tab in the Settings form. If the component doesn't have Settings, Kiln will generate this too.
--- a/docs/components.md
+++ b/docs/components.md
@@ -57,3 +57,7 @@ Models must be compiled and added to `window.kiln.componentModels` for Kiln to c
 Schemas \(or _schemata_, for pedants like myself\) are the main connective tissue between components and Kiln. They define what kinds of fields components have, as well as how those fields are grouped. They also provide `_description` text for the components' Info Modals, `_version` for upgrading data programmatically, and `_confirmRemoval` for extra peace-of-mind on components like ads.
 
 Schemas for components that are currently on the page are automatically parsed by Amphora and added to the page when Kiln loads. Schemas for other components are added when those components are added.
+
+## Variations
+
+Components can have variations, which is defined by corresponding stylesheets in the site's styleguide folder. There is the default variation (`componet-name.css`)and variations are defined as `component-name_variation.css`. Kiln will automatically generate a select input with the variations and a 'Component Variation' tab in the Settings form. If the component doesn't have Settings, Kiln will generate this too.
--- a/inputs/select.vue
+++ b/inputs/select.vue
@@ -123,12 +123,16 @@
       // combine arg/prop options, fetched list options, and a null option for non-multiple selects
       options() {
         const propOptions = this.args.options || [],
-          currentSlug = _.get(this.$store, 'state.site.slug');
+          currentSlug = _.get(this.$store, 'state.site.slug'),
+          noneOption = propOptions.filter((val)=>{
+            return val.name === 'None';
+          });
 
         let fullOptions = propOptions.concat(this.listOptions);
 
-        // single-select must have null option
-        if (fullOptions.length && !this.args.multiple) {
+        // if there is no 'None' option defined then add a null option because
+        // single-select must have null option, if there is no null option
+        if (fullOptions.length && !this.args.multiple && !noneOption.length) {
           fullOptions = [this.NULL_OPTION].concat(fullOptions);
         }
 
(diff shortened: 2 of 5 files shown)


inputs/select.vue
```diff
@@ -123,12 +124,16 @@
       // combine arg/prop options, fetched list options, and a null option for non-multiple selects
       options() {
         const propOptions = this.args.options || [],
-          currentSlug = _.get(this.$store, 'state.site.slug');
+          currentSlug = _.get(this.$store, 'state.site.slug'),
+          noneOption = propOptions.filter((val)=>{
```
reubenson: find would be more performant than filter here?

inputs/select.vue
```diff
@@ -123,12 +124,16 @@
...
+            return val.name === 'None';
```
reubenson: should default to 'Default' instead of 'None'?

docs/components.md
```diff
@@ -57,3 +57,7 @@ Models must be compiled and added to `window.kiln.componentModels` for Kiln to c
 Schemas \(or _schemata_, for pedants like myself\) are the main connective tissue between components and Kiln. They define what kinds of fields components have, as well as how those fields are grouped. They also provide `_description` text for the components' Info Modals, `_version` for upgrading data programmatically, and `_confirmRemoval` for extra peace-of-mind on components like ads.
 
 Schemas for components that are currently on the page are automatically parsed by Amphora and added to the page when Kiln loads. Schemas for other components are added when those components are added.
+
+## Variations
+
+Components can have variations, which is defined by corresponding stylesheets in the site's styleguide folder. There is the default variation (`componet-name.css`) and variations are defined as `component-name_variation.css`. Kiln will automatically generate a select input with the variations and a 'Component Variation' tab in the Settings form. If the component doesn't have Settings, Kiln will generate this too.
```
cperryk: typo `(`componet-name.css`)`. Also not sure what you mean by this: `If the component doesn't have Settings, Kiln will generate this too.`

inputs/select.vue
reubenson: nvm

inputs/select.vue
```diff
@@ -123,12 +124,16 @@
       // combine arg/prop options, fetched list options, and a null option for non-multiple selects
       options() {
         const propOptions = this.args.options || [],
-          currentSlug = _.get(this.$store, 'state.site.slug');
+          currentSlug = _.get(this.$store, 'state.site.slug'),
+          hasNoneOption = propOptions.find((val)=>{
+            return val.name === 'None';
+          });
```
cperryk: `propOptions.some` would be clearer

inputs/select.vue
```diff
@@ -123,12 +124,16 @@
       // combine arg/prop options, fetched list options, and a null option for non-multiple selects
       options() {
         const propOptions = this.args.options || [],
-          currentSlug = _.get(this.$store, 'state.site.slug');
+          currentSlug = _.get(this.$store, 'state.site.slug'),
+          hasNoneOption = propOptions.find((val)=>{
```
reubenson: `val => val.name === 'None` for brevity?

lib/core-data/get-variations.js
```diff
@@ -10,23 +10,35 @@ const log = logger(__filename);
  * @returns {object}
  **/
 function addField(schema, componentVariations) {
-  const variations = _.map(componentVariations, function (variant) {
+  // making the assumption that the component variation name is prefixed by the
+  // component name
+  const componentName = componentVariations[0].split('_')[0],
```
reubenson: what happens if/when this errors?

docs/components.md
```diff
@@ -57,3 +57,7 @@ Models must be compiled and added to `window.kiln.componentModels` for Kiln to c
 Schemas \(or _schemata_, for pedants like myself\) are the main connective tissue between components and Kiln. They define what kinds of fields components have, as well as how those fields are grouped. They also provide `_description` text for the components' Info Modals, `_version` for upgrading data programmatically, and `_confirmRemoval` for extra peace-of-mind on components like ads.
 
 Schemas for components that are currently on the page are automatically parsed by Amphora and added to the page when Kiln loads. Schemas for other components are added when those components are added.
+
+## Variations
+
+Components can have variations, which is defined by corresponding stylesheets in the site's styleguide folder. Variations are defined as `component-name_variation.css`. Kiln will automatically generate a select input with the variations. This input will be in the 'Component Variation' tab in the Settings group. If the component doesn't have a Settings group set, Kiln will generate this too.
```
reubenson: ✨

lib/core-data/get-variations.js
amycheng: `amphora-html` depends on this convention to figure out what component has variations, it then passes this data to Kiln. If this convention is broken, then that component will not have variations and this logic will not be run.

docs/components.md
nelsonpecora: I'd change the example to `component-name_variation-name.css`, to show that variations are also hyphen-separated.

inputs/select.vue
```diff
@@ -35,7 +35,8 @@
 
   > #### info::Usage Notes
   >
-  > * no/empty option is pre-selected by default (you don't need to specify an empty option in the schema)
+  > * you can have a value for a 'None' option, this is useful for components that have defaults that you want to be able to revert to after selecting an option
```
nelsonpecora: I would add an example showing this in action, since it's a little confusing at first glance

lib/core-data/get-variations.js
nelsonpecora: What happens if there are no variations in the store (e.g. this version of Kiln is running against an older, non-variation version of amphora)? If it's taken into account already, please add a comment to that effect.

inputs/select.vue
amycheng: @nelsonpecora Is this what you meant? https://github.com/clay/clay-kiln/pull/1161/files#diff-15e2547214bfb6b1eba657f0bb1b5c91R50

nelsonpecora: yes!

PR #1164, opened 2018-04-09
complex-list reordering

nelsonpecora:
* FEATURE: ability to reorder complex list items!
* item's position in the list will now always display, with buttons to move it up / down
* when tabbing through fields (and buttons) in list items, tabbing to something in another list item will select that item (thus displaying the remove/add buttons)
* updated styling and animation for the bottom of list items to reflect the new affordances

### DIFF
--- a/inputs/complex-list-item.vue
+++ b/inputs/complex-list-item.vue
@@ -8,7 +8,7 @@
     border: 1px solid $divider-color;
     border-radius: 2px;
     box-shadow: $shadow-2dp;
-    padding: 16px;
+    padding: 16px 16px 0;
     transition: all 200ms $standard-curve;
 
     &:hover {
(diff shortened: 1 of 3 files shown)


inputs/complex-list-item.vue
```diff
@@ -33,13 +33,17 @@
   }
 
   .complex-list-item-actions {
-    margin-top: 10px;
-    width: 100%;
+    border-top: 1px solid $card-border-color;
+    margin: 16px -16px 0;
+    // account for outer padding
+    padding: 8px 16px;
+    width: calc(100% + 32px);
```
cperryk: Could this be handled instead with `box-sizing: border-box`?

nelsonpecora: It's already using border-box, I believe

PR #1172, opened 2018-04-13
Fixes timepicker component not populating value (#1157)

macgyver:
* Fixes parsing of stored value in timepicker

* Do not parse stored value

* Simplify

### DIFF
--- a/lib/utils/timepicker.vue
+++ b/lib/utils/timepicker.vue
@@ -52,7 +52,7 @@
       }
     },
     mounted() {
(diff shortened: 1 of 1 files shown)


macgyver:
Hey @nelsonpecora just trying to get this train moving, we've got kiln-5 in staging and we are hoping to go to prod early next week. It would be really nice to get a new minor version of kiln 5 with this bugfix available on npm soon. 

I'm not sure about the failing tests - it looks like an npm dependency issue on Circle CI, but I don't have permissions to configure any of that stuff. Let me know if it's something we can help with on our end..

nelsonpecora:
Hmm, I'll manually re-trigger the build and see if that helps.

PR #1193, opened 2018-04-25
collapsible complex-list items

nelsonpecora:
* complex-list now supports collapsible items!
* you can make a complex-list collapsible with the `collapse` argument
* specify what property you want to display as the title of the item when it's collapsed (if that property is empty, it'll say "New Item")
* only one item may be expanded at a time. this is the current item! (same item that gets the action buttons at the bottom)

### DIFF
--- a/inputs/complex-list-item.vue
+++ b/inputs/complex-list-item.vue
@@ -166,7 +166,7 @@
         return _.has(this.schema, 'collapse');
       },
       collapsedTitle() {
-        return _.get(this.data, _.get(this.schema, 'collapse'), 'New Item');
+        return _.get(this.data, _.get(this.schema, 'collapse')) || 'New Item';
       },
       props() {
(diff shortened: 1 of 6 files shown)


inputs/complex-list-item.vue
```diff
@@ -148,18 +200,18 @@
...
         // true if any of the fields in the current item have required validation
-        return _.some(this.schema, (obj) => _.has(obj, `${fieldProp}.validate.required`));
+        return _.some(this.props, (obj) => _.has(obj, `${fieldProp}.validate.required`));
       }
     },
     methods: {
       enter(el, done) {
         this.$nextTick(() => {
-          velocity(el, 'fadeIn', { duration: 375, complete: done });
+          velocity(el, 'fadeIn', { duration: 250, complete: done });
```
byronhulcher: 100% here for shorter transition times

PR #1208, opened 2018-05-09
disallow scheduling in the past

### LINKED ISSUE #1205: prevent scheduling in the past
Currently, we disable the "schedule" button if the date and time aren't both filled out, and it's impossible to select a date in the past, but it's possible to select a _time_ in the past (which would force the page to publish immediately). We should close this loophole.

nelsonpecora:
* disable the schedule button when the date/time are set to sometime in the past
* tested against keyboard input (hitting enter in that form) as well
* fixes #1205

### DIFF
--- a/lib/drawers/publish.vue
+++ b/lib/drawers/publish.vue
@@ -200,6 +200,7 @@
   import addWeeks from 'date-fns/add_weeks';
   import subWeeks from 'date-fns/sub_weeks';
   import isThisWeek from 'date-fns/is_this_week';
+  import isPast from 'date-fns/is_past';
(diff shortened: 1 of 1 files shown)


lib/drawers/publish.vue
```diff
@@ -237,6 +238,14 @@
     }
   }
 
+  function isInThePast(dateValue, timeValue) {
```
cperryk: Can we add code comments that make it clear what `dateValue` and `timeValue` are?

nelsonpecora: good point

PR #1213, opened 2018-05-11
fetch styles when adding new components

nelsonpecora:
* updates component creation to support new styleguides
* refactors legacy style fetching to use `<link>` tags rather than inlining styles (matching how edit mode works nowadays)
* supports both site styleguides and `_default` styleguide, as well as variations in both
* only loads `_default` styles if site styleguide does not contain styles for specified component

### DIFF
--- a/lib/component-data/create.test.js
+++ b/lib/component-data/create.test.js
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import * as store from '../core-data/store';
 import * as components from '../core-data/components';
 import * as api from '../core-data/api';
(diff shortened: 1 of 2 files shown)


lib/component-data/create.test.js
```diff
@@ -94,18 +93,6 @@ describe('component creator', () => {
     });
   });
 
-  it('fetches base component styles for new components', () => {
```
nelsonpecora: because we're just adding link tags, there's not much to mock without mucking with the "global" store, so I'm not sure if this can be unit tested effectively

Issue #1189, opened 2018-04-24
Feature request: custom transform functions for magic buttons


macgyver:
Use case: We have a different url slugification strategy than nymag, it would be nice to define our own custom transform function in our "sites" repo, similar to how the plugins for inputs, modals, etc.. work

nelsonpecora:
This would be super useful!

nelsonpecora:
note: let's do #1219 (trimming urls in the url transformer) as part of this

PR #1219, opened 2018-05-21
Magic button trim

KerenLerner:
Magic button should trim leading and trailing spaces from a string before creating a url.  Otherwise, it will put dashes at the beginning and end of a slug whose input string contains extra space (e.g. `mysite.com/-weather-`)


nelsonpecora:
Thanks for catching this! Adding it to https://github.com/clay/clay-kiln/issues/1189 since that'll touch the same code

PR #1223, opened 2018-05-22
global transformers api

### LINKED ISSUE #1189: Feature request: custom transform functions for magic buttons

nelsonpecora:
* adds built-in transformers to `window.kiln.transformers`
* updates docs, adds all properties of `window.kiln` with descriptions
* fixes #1189


byronhulcher:
Thanks for adding more docs while you were in there!

PR #1230, opened 2018-05-30
Layout Editing

nelsonpecora:
## FEATURE: Toggleable Layout Editing

Kiln is now split into two edit modes, *Page Editing* and *Layout Editing*. The former is very similar to the current way most users use Kiln, namely creating, editing, and publishing page-specific content. The latter mode is restricted to admins, and allows them to edit and publish content in the layout. This is intended to prevent non-admins from messing with layout-level components (which affect upwards of thousands of pages for some layouts), and creates enough friction for admins that they will also (hopefully) not edit layouts unintentionally.

**Page Editing Changes:**

* Users can manually set the Kiln page title (the title that appears in the Clay menu) from a section in the Publishing drawer
* The top right toolbar button now says "Publishing," rather than using different verbs (Publish, Unpublish, etc) depending on the page state
* Placeholders and buttons using the accent color are now light blue (formerly indigo), allowing them to stick out better
* Permanent placeholders are now blue-grey (formerly light grey), differentiating them from placeholders in the inactive edit mode
* When hovering over a component in the inactive edit mode, a small tooltip will appear rather than the component selector
* Non-admins will not see the option to enter Layout Edit Mode
* Publishing / Scheduling the page will _not_ trigger publishing / scheduling of the layout
* Find on Page will only show components in the page (and only include tabs for head / invisible component lists that are page-specific), removing some clutter
* Pre-publish validation will only run on page-specific components

slack-imgs

**Layout Editing Changes:**

* Admins will see three options in the toolbar: "Edit Page," "Edit Layout," and "View Page"
* When entering Layout Edit Mode, users will see an alert informing them that they're editing the layout (this is a similar warning to the one displayed when editing a New Page Template)
* When editing a layout that was recently edited by someone else, users will see an alert (similar to the alert they see when editing a page that was recently edited by someone else)
* Find on Layout will only show components in the layout (and only include tabs for head / invisible component lists that are layout-specific)
* The simplified Layout History drawer will show the history of changes to the layout
* The layout Publishing drawer allows users to schedule, unschedule, publish, and change the title of layouts
* Pre-publish validation will only run on layout-specific components

slack-imgs-1

**Notes:**

* Layouts must be denoted with a `_layout: true` root-level property in their schemas
* This feature requires `amphora-search` `v6.2.0` or above to function
* This also fixes an issue where the Clay avatar was not being displayed for history actions triggered by amphora (e.g. automatically publishing scheduled content)
* In the future, we may add more affordances for creating and managing layouts, including seeing which layout a page uses in the Clay menu

slack-imgs-2


lib/deep-linking/actions.js
```diff
@@ -36,6 +45,23 @@ function openHashedForm(store, urlProps) {
...
 
   if (find(`[${refAttr}="${uri}"]`) && el) {
+    // el is in the body, not the head. check to see if we have permission to edit it
+    const isPageComponent = isComponentInPage(uri);
+
+    if (!isPageComponent && !isAdmin) {
+      // layout component that cannot be edited! display a warning to the user
+      const name = label(getComponentName(uri));
+
+      store.dispatch('addAlert', { type: 'error', text: `You do not have permission to edit this "${name}". Please contact an admin for assistance.` });
```
byronhulcher: Is `admin` a concept a writer/user would recognize? Would they know this means an editor? Or would they just go into #help

lib/drawers/publish-page.vue
```diff
@@ -179,6 +198,15 @@
       </form>
     </ui-collapsible>
 
+    <!-- manual title updating -->
+    <ui-collapsible class="publish-section publish-title" title="Page Title">
+      <form class="publish-title-form" @submit.prevent="saveTitle">
+        <span class="title-description">Manually update the page title that appears in the Clay Menu. Will be overwritten when saving components that set the title.</span>
```
byronhulcher: I don't like the wording of this (I dont think writers or editors know what does/doesn't update the title)

lib/utils/avatar.vue
```diff
@@ -32,8 +32,14 @@
         }
       },
       imageURL() {
-        // only display the image if it's not the default google avatar
-        if (this.url && this.url.length && this.url !== 'https://lh3.googleusercontent.com/-XdUIqdMkCWA/AAAAAAAAAAI/AAAAAAAAAAA/4252rscbv5M/photo.jpg?sz=50') {
+        // only display the image if it's not the default google avatar (or the built-in clay avatar)
+        if (this.url && this.url === 'kiln-clay-avatar') {
+          // special kiln avatar
+          // note: we're referencing it from amphora-search instead of hardcoding it so we can
```
byronhulcher: isn't this... hardcoding it???

lib/utils/avatar.vue
```diff
@@ -32,8 +32,14 @@
...
+        } else if (this.url && this.url.length && this.url !== 'https://lh3.googleusercontent.com/-XdUIqdMkCWA/AAAAAAAAAAI/AAAAAAAAAAA/4252rscbv5M/photo.jpg?sz=50') {
```
byronhulcher: what

lib/decorators/selector.vue
```diff
@@ -201,6 +221,7 @@
         <ui-icon-button v-show="hasAddComponent && !hasAddSingleComponent && isBelowMaxLength" type="secondary" color="primary" class="quick-bar-button quick-bar-add" icon="add" :tooltip="addComponentText" @click.stop="openAddComponentPane"></ui-icon-button>
       </div>
     </aside>
+    <aside v-else-if="isCurrentSelection && !isActive" class="mini-selector-inactive">Please Use {{ activeModeLabel }} Mode to Edit</aside>
```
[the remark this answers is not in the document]
nelsonpecora: `isCurrentSelection && !isActive`...

lib/deep-linking/actions.js
```diff
@@ -1,15 +1,24 @@
...
 
 function openHashedForm(store, urlProps) {
-  let component, instance, path, foundComponent, uri, el, group, groupEl;
+  let isAdmin, component, instance, path, foundComponent, uri, el, group, groupEl;
+
+  // get user permissions
+  // note: as clay always opens pages in page edit mode (not layout),
+  // we'll have to switch modes if the deep link is to a layout component
+  // (and the user has permission to edit it)
+  isAdmin = _.get(store, 'state.user.auth') === 'admin';
```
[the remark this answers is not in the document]
nelsonpecora: It's short enough that I'm comfortable doing it in here, because the alternative is passing the whole `state` object around

lib/deep-linking/actions.js
nelsonpecora: Not entirely sure. I think some of them (copyeditors) are familiar, but I'm fine if they go into #help. Also the wording will definitely be tweaked as we get user feedback.

lib/drawers/publish-page.vue
nelsonpecora: The wording will be tweaked as we get feedback, but I agree

lib/utils/avatar.vue
nelsonpecora: "in amphora-search, we're referencing <giant blob of base64-encoded image> rather than hardcoding it"

lib/utils/avatar.vue
nelsonpecora: welcome to gooooogle
they're so swell
their default avi's that url
you think it's odd?
it's weird as hell

PR #1232, opened 2018-05-30
small kiln tweaks

nelsonpecora:
* update beyonce validator to not run on url inputs
* fix bug with meta keys on firefox
* fix bug with min/max validation on number inputs
* change 'None' to 'Default' in variation selector
* add default variation to component data when rendering client-side (if it isn't set)

### DIFF
--- a/edit.js
+++ b/edit.js
@@ -192,7 +192,7 @@ document.addEventListener('DOMContentLoaded', function () {
   // undo and redo with shortkey+z / shift+shortkey+z (e.g. Ctrl+z, Shift+Ctrl+z)
   // display cheat sheet of all keyboard shortcuts with shift+?
   // toggle the meta key with ctrl / left command
-  /* eslint-disable */
+  /* eslint-disable complexity */
   document.body.addEventListener('keydown', (e) => {
     const key = keycode(e),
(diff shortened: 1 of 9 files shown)


edit.js
```diff
@@ -91,6 +93,19 @@ function getLastEditUser(store) {
   return lastUser ? lastUser.name : null;
 }
 
+/**
+ * determine if forms, or modals are open
+ * @param  {object}  store
+ * @return {Boolean}
+ */
+function isStuffOpen(store) {
```
byronhulcher: tfw the naming is too hard

edit.js
```diff
@@ -173,10 +188,34 @@ document.addEventListener('DOMContentLoaded', function () {
   });
 
   // when ESC bubbles up to the document, close the current form or pane / unselect components
+  // navigate components when hitting ↑ / ↓ arrows (if there's a component selected)
+  // undo and redo with shortkey+z / shift+shortkey+z (e.g. Ctrl+z, Shift+Ctrl+z)
+  // display cheat sheet of all keyboard shortcuts with shift+?
+  // toggle the meta key with ctrl / left command
+  /* eslint-disable */
```
byronhulcher: Please only disable rules that this code breaks

PR #1239, opened 2018-06-08
add new _reveal functionality, fix bugs

nelsonpecora:
* FEATURE: add `_reveal` to `segmented-button-group` options
* FEATURE: allow `_reveal` to match against multiple `values`
* fixed a lot of bugs related to form resizing, especially when using `complex-list` and/or `_reveal`
* updated docs for new features

### DIFF
--- a/lib/forms/field-helpers.js
+++ b/lib/forms/field-helpers.js
@@ -5,6 +5,7 @@ import striptags from 'striptags';
 import { closest, find } from '@nymag/dom';
 import { fieldClass, firstFieldClass } from '../utils/references';
 import { isEmpty, compare } from '../utils/comparators';
+import { filterBySite } from '../utils/site-filter';
 import logger from '../utils/log';
 
 const log = logger(__filename);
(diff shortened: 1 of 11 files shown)


lib/forms/field-helpers.js
```diff
@@ -207,3 +208,67 @@ export function getValidationError(data, validate, store, name) { // eslint-disa
...
+
+/**
+ * determine if something should be revealed based on the sites specified
+ * and the current site
+ * @param  {string} sites
+ * @param  {string} currentSlug
+ * @return {boolean}
+ */
+function shouldBeRevealedBySite(sites, currentSlug) {
+  return !!filterBySite([{ sites }], currentSlug).length;
```
dmabuada: this might be a future feature request, but it might be useful to have this filterable by more than just the site (e.g. i had a need for nymag site, but subscription pages only: `nymag.com/magazine/subscription`)

nelsonpecora: Ah, that might be useful. I think we'd have to do some kind of discovery though, since I'm not sure what the scope of use cases would be.

PR #1242, opened 2018-06-11
fix issues with falsy values, object comparison

nelsonpecora:
* fixes issue when comparing against a falsy `value`
* fixes issues when comparing objects and arrays (`===` comparator now runs `_.isEqual` which checks equality by value)

### DIFF
--- a/lib/forms/field-helpers.js
+++ b/lib/forms/field-helpers.js
@@ -229,14 +229,12 @@ function shouldBeRevealedBySite(sites, currentSlug) {
  * @return {boolean}
  */
 function shouldBeRevealedByData(data, operator, value, values) {
-  if (value) {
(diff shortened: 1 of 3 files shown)


lib/forms/field-helpers.js
```diff
@@ -229,14 +229,12 @@ function shouldBeRevealedBySite(sites, currentSlug) {
  * @return {boolean}
  */
 function shouldBeRevealedByData(data, operator, value, values) {
-  if (value) {
+  if (value !== undefined) {
```
byronhulcher: I think `typeof value !== "undefined"` is the canonical way to do this

nelsonpecora: oh right, because `undefined` can be overwritten. <3 javascript

Issue #1247, opened 2018-06-13
For array inputs, automatically set validate.required if validate.min = 1 is set


amycheng:
## Idea
For inputs that are arrays like `complex-list`, `validate.min = 1` will not trigger if the field is empty. To get the desired behavior, the developer has to set `validate.required = true`. `validate.min = 1` and `validate.required = true` are essentially the same for array inputs.

## Proposed Solution/Implementation
If one sets `validate.min = 1`, then `validate.required = true` could be automatically set to avoide confusion.

nelsonpecora:
Keeping them separate is by design. Requiring a minimum value _if a field is set_ is different than requiring the field to be set, and that holds the same across all data types.

PR #1249, opened 2018-06-14
Fix min validation for complex-list when is empty

CharlieGarcia:
There was an issue where complex-list components were specified a minimum amount of fields were been checked as valid and allowing the page to be published.

```yml
validate:
      min: 1
      max: 4
```


nelsonpecora:
Per [this comment](https://github.com/clay/clay-kiln/issues/1247#issuecomment-397099594), the distinction between `required` and `min` validation is by design. You can have apply both `required` and `min` validation to fields that support them, but `min: 1` is _not_ equivalent to `required: true`

PR #1259, opened 2018-06-25
new filterable lists

### LINKED ISSUE #1016: clean up filterable lists
* [x] refactor lists to use `$emit` (similarly to `person` components)
* [x] update hover states for list item buttons
* [x] use slots for list items / actions, to allow for custom buttons

### LINKED ISSUE #1169: Named Instances
* ability to label and save certain instances of components to a list
* ability to select from that list when adding that type of component (rather than simply cloning the default data)
* note: you'd still be cloning the data (not just referencing the instance), so no worries around scopes / components updating in unknown places

named list

Questions:

* who can add and remove named instances from the list?
* how does someone _edit_ the instances in the list? (do they reference the original, or do they get cloned when being added to the list and exist on their own?)
* who is allowed to add from this list?
* how does someone add from the list? (idea in the drawing above)

### LINKED ISSUE #1251: Optimize New Pages List
Breaking this out from https://github.com/clay/clay-kiln/issues/1165

We should optimize the New Pages List by adding categories and having a consistent sorting:

* when adding new page templates, specify a category (using autocomplete / autosuggest to reduce clerical errors)
* sort New Pages List alphabetically by category and page title
* pages _without_ a category will automatically be put into a "General" category (solves migration issues)
* categories will be dropdowns, per [Expand and Collapse](https://material.io/design/components/lists.html#types)
* filtering will be supported. filtering will automatically open categories that match (or have items that match) the filter
* keyboard selecting will be supported. up/down arrows will navigate categories and items. right arrow will open/close items (enter will execute the root action if it exists, otherwise it'll also open/close items)
todo:

* [x] should categories have icons?
* [x] doesn't depend on new `/_lists` implementation, but would be nice to have

nelsonpecora:
* fix #1016, makes filterable lists more consistent
* fix #1251, optimize new pages list
* fix #1169, named component instances

This PR has _quite_ a few cool things, all centered around filterable lists:

* BREAKING: filterable list api has been changed to use proper Vue.js events, instead of passing functions into list items
* BREAKING: filterable list api has been changed so you may pass in whatever "secondary actions" you desire
* FEATURE: filterable lists may contain items that expand/collapse
* FEATURE: all "New Page Templates" are now in categories, and admins can determine the category when adding pages to that list
* FEATURE: certain component instances can be "bookmarked" (add `_allowBookmarks: true` in the schema), allowing admins to save specific named component instances. These will appear when adding components
* fixed keyboard support across all filterable lists
* updated screenshots in docs
* added `FINISHED_DECORATING` vanity mutation (for plugins)

## New Filterable List API

**Properties:**

* `content` - array of objects with `id`, `title`, and optionally `children`
* `secondaryActions` - array of objects with `icon`, `tooltip`, `action` (that happens on click), `enable` (that determines if the button should display) (`icon` and `tooltip` may be functions that get `id` of the items as their first argument)
* `header` - string with a header title. setting this will display column headers in the list
* `addTitle` - string with a custom title for the add button
* `addIcon` - string with a custom icon for the add button
* `filterLabel` - string with a custom label for the filter input
* `filterHelp` - string with custom help text for the filter input
* `initialExpanded` - `id` of the root-level item that should be expanded when the list is displayed

**Events:**

* `@reorder` - will be called when items are reordered. adding this listener will make the list reorderable, but prevent filtering. Note that reordering items with children is currently unsupported
* `@root-action` - called when root-level items are clicked, or `enter` has been pressed after they're navigated to via the keyboard
* `@child-action` - similar to `@root-action`, but called on child items. adding this listener will enable expanding/collapsing of root items (if they have children)
* `@add` - called when add button is clicked. adding this listener will display the add button

## Examples

**New Page Templates:**

screen shot 2018-06-25 at 4 33 10 pm

Filtering by the name of a root-level item will display all children:

screen shot 2018-06-25 at 4 34 44 pm

When adding a page template, categories will be autosuggested:

screen shot 2018-06-25 at 4 34 14 pm

**Component Bookmarks**

screen shot 2018-06-25 at 4 36 27 pm

screen shot 2018-06-25 at 4 36 52 pm

Admins can remove bookmarks from the list:

screen shot 2018-06-25 at 4 38 09 pm

Filtering by child items will only display relevant matches:

screen shot 2018-06-25 at 4 37 53 pm


package.json
```diff
@@ -168,5 +168,6 @@
   },
   "peerDependencies": {
     "amphora-search": "^6.2.0"
-  }
+  },
+  "dependencies": {}
```
reubenson: needed?

nelsonpecora: eh, it's just npm doing its thing 🙃

PR #1267, opened 2018-07-03
support comparing against multiple values in conditional required validation

zonika:
The kiln guide says that validate.required supports the same properties as `_reveal` to decide whether or not a field is required. This isn't exactly true as validate.required wasn't looking for a `values` array to compare against. Now it is.


nelsonpecora:
or v7.1.0 I guess

Issue #1258, opened 2018-06-25
Trigger Render of Editing Modal on Complex List Add / Remove


yuliyv:
The way kiln currently handles `_reveal` logic with `complex-list` inputs means it's challenging to build editing experiences that react to the addition / removal of elements from the list.

Consider the following sample schema file:

```yml
description: |
  Sample Component

collection:
  _has:
    input: complex-list
    props:
      -
        prop: data
        _label: Data
        _has:
          input: text
          help: Some sample data

collectionLayout:
  _label: Sample Collection Layout
  _has:
    input: text
  _reveal:
    field: collection.length
    operator: '>'
    value: 0

_groups:
  settings:
    fields:
      - collection
      - collectionLayout
```

In this example we see that the component has a property `collectionLayout` which is conditionally rendered based on the length of the `collection` complex list.  With the current way Kiln works this `_reveal` logic is not triggered when elements are added/removed from the `collection` complex list so a user must close the editing modal and re-open it before the `_reveal` logic is executed.  Ideally instead this `_reveal` logic would be triggered as soon as elements are added / removed from the list

nelsonpecora:
Hmm, we don't currently support that `collection.length` field property, which is why it's probably only being run when it's checking the store rather than the current field. Checking lengths is an interesting feature request, but it looks like in this instance you really only care if the list is not empty, correct? You can do that with the `not-empty` operator (no value required).

yuliyv:
This is perhaps a too simplified example, here's a more robust example of what I am trying to achieve:

```yml
description: |
  Sample Component

collection:
  _has:
    input: complex-list
    props:
      -
        prop: data
        _label: Data
        _has:
          input: text
          help: Some sample data url

collectionLayout:
  _label: Sample Collection Layout
  _has:
    input: segmented-button-group
    options:
      - title: 2 Images
        values:
          - icon: icon-thumbnail.svg
            text: Two Vertical
            value: two-vertical
          - icon: icon-thumbnail.svg
            text: Two Square
            value: two-square
        _reveal:
          field: collection.length
          operator: '==='
          value: 2
      - title: 3 Images
        values:
          - icon: icon-thumbnail.svg
            text: Three Vertical
            value: three-vertical
          - icon: icon-thumbnail.svg
            text: Three Square
            value: three-square
        _reveal:
          field: collection.length
          operator: '==='
          value: 3
  _reveal:
    field: collection.length
    operator: '>'
    value: 0

_groups:
  settings:
[...]

PR #1273, opened 2018-07-09
Add 'size-of' Comparator

yuliyv:
Related to the outcome of https://github.com/clay/clay-kiln/issues/1258 this PR adds a new comparator `size-of` which is useful for triggering schema `_reveal` logic based on the size of a complex list.  Here's an example of a how this could be used:

```yml
description: |
  Sample Component

collection:
  _has:
    input: complex-list
    props:
      -
        prop: data
        _label: Data
        _has:
          input: text
          help: Some sample data url

collectionLayout:
  _label: Sample Collection Layout
  _has:
    input: segmented-button-group
    options:
      - title: 2 Images
        values:
          - icon: icon-thumbnail.svg
            text: Two Vertical
            value: two-vertical
          - icon: icon-thumbnail.svg
            text: Two Square
            value: two-square
        _reveal:
          field: collection
          operator: 'size-of'
          value: 2
      - title: 3 Images
        values:
          - icon: icon-thumbnail.svg
            text: Three Vertical
            value: three-vertical
          - icon: icon-thumbnail.svg
            text: Three Square
            value: three-square
        _reveal:
          field: collection
          operator: 'size-of'
          value: 3
  _reveal:
    field: collection
    operator: 'size-of'
    value: 0

_groups:
  settings:
    fields:
      - collection (Items)
      - collectionLayout (Item Layout)
```

With this operator it will now trigger the `_reveal` actions as makes changes and switches tabs.

### DIFF
--- a/docs/editing-components.md
+++ b/docs/editing-components.md
@@ -244,7 +244,7 @@ When comparing against multiple `values`, the comparison will be true if at leas
 * `>=`
 * `typeof`
 * `regex`
-* `size-of`
+* `length-equals`
 * `empty` (only checks field data, no value needed)
 * `not-empty` (only checks field data, no value needed)
 * `truthy` (only checks field data, no value needed)
--- a/docs/editing-components.md
+++ b/docs/editing-components.md
@@ -244,6 +244,7 @@ When comparing against multiple `values`, the comparison will be true if at leas
 * `>=`
 * `typeof`
 * `regex`
+* `size-of`
 * `empty` (only checks field data, no value needed)
 * `not-empty` (only checks field data, no value needed)
 * `truthy` (only checks field data, no value needed)
--- a/docs/editing-components.md
+++ b/docs/editing-components.md
@@ -244,7 +244,9 @@ When comparing against multiple `values`, the comparison will be true if at leas
 * `>=`
 * `typeof`
 * `regex`
-* `length-equals`
+* `length-equals` (compares the length of strings and arrays)
+* `length-less-than` (compares the length of strings and arrays)
+* `length-greater-than` (compares the length of strings and arrays)
 * `empty` (only checks field data, no value needed)
 * `not-empty` (only checks field data, no value needed)
 * `truthy` (only checks field data, no value needed)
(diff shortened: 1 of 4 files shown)


lib/utils/comparators.js
```diff
@@ -12,7 +12,8 @@ const operators = {
   empty: (l) => isEmpty(l),
   'not-empty': (l) => !isEmpty(l),
   truthy: (l) => !!l,
-  falsy: (l) => !!!l
+  falsy: (l) => !!!l,
+  'size-of': (l, r) => _.size(l) === r
```
byronhulcher: Pedantic but I think it should be `size-equals` and maybe even `length-equals`.  I know you're using lodash's size function but i dont think we'd ever plan to use this with an object would we? so its really just checking `(l, r) => l.length === r`

yuliyv: I am definitely fine with changing this to not utilize lodash, I used it here to have the additional benefits of automatically handling things which are not 'object-like'.  I think if I change to the above it would also benefit from some light type checking.  Will update the name as well since it is explicitly checking lengths

zonika: I definitely feel like if we arent type checking the input we should be using lodash here. That being said, do we feel like there is a use case for this to work for strings, etc? If so, using lodash _.size feels like the right call as it handles arrays, objects, and strings.

byronhulcher: I'm totally cool with lodash i think size is better for type safety., i was more referencing its use discussing `size-equals` vs `length-equals` for the name

yuliyv: Sorry misread your comment, sure yeah it makes sense to just alter the name

yuliyv:
@zonika Yeah it will also work with conditional requires

docs/editing-components.md
```diff
@@ -244,6 +244,7 @@ When comparing against multiple `values`, the comparison will be true if at leas
 * `>=`
 * `typeof`
 * `regex`
+* `length-equals`
```
nelsonpecora: Yep, I'd change the name to `length-equals` (since "length" is used for other things instead of "size" and so the configs read better). If you're already here it might be nice to also add `length-greater-than` and `length-less-than` since I could easily see them being used.

Plus the docs should say that it compares the length of strings and arrays

PR #1277, opened 2018-07-09
add focus to deep-linking

### LINKED ISSUE #1266: Allow ui elements inside modals to open additional modals as stack vs singleton
The problem currently arises as a result of NY Mag's `mediaplay-picker` kiln plugin's interaction with a complex list inside of a kiln group inside of a modal.  The `mediaplay-picker` plugin is a button which can be appended to inputs which opens a modal where a user can select one or more photos.  When closing this modal using the modal's built-in close/`X` button, or by resolving the modal by selecting and inserting images, the modal closes, returning the user to the page.  If this sequence is triggered by an input inside of a kiln modal, it still returns the user to the page, rather than to the input inside of the modal. 

modal

The new `image-collection` component uses a complex list, and one of the inputs of this complex list will have the `mediaplay-picker` option. I'd like to be able to pass an argument to when I open the modal to add it to a stack instead of replacing the existing modal, so when I close out of the second modal I'm returned to the first with the complex list, so that I can continue editing.

### LINKED ISSUE #1272: More consistent focus on the first field of forms
When a settings form modal is opened, and complex list is on that form with an empty input field as the first input for the first item, I'd like that input field recieve focus, as an empty text input outside of a complex list would if it was the first item in the settings form.

For settings forms with  an empty text input as the first element:
```
1) I click the settings gear or placeholder to open a settings form
2) The first panel is opened, revealing an empty text input
3) The first input in the form is focused
```

For settings forms with a complex-list containing a single empty item as the first element
```
1) I click the settings gear or placeholder to open a settings form
2) The first panel is opened, revealing a complex list with a single item with all empty fields

Expected:
3) The first input for the first item should be focused

Actual:
3) No inputs have focus
```

nelsonpecora:
* fixes #1272 
* fixes #1266 

**FEATURE:** Add individual field focus to the deep linking system

This allows deep-linking to individual fields, rather than just linking to forms. When a form is opened, the relevant field will be focused (and the section its in will be switched to, when applicable). This is backwards compatible with our current deep-linking, and fixes a number of issues with field autofocus (if you aren't explicitly focusing a field, the first field in the form will now be focused, which also works with complex-lists).

Deep links with focus look like this (note: this is linking to a specific item inside a complex-list):

```
#image~cjjess29h00133i64jmqkhf08~settings~content.1.imageUrl
```

One side effect of this is that clicking validation issues will now go to the correct section tab and focus on the relevant field, rather than making the end-user click around a complex form looking for it.

Also fixed a bug where site-specific fields weren't opening the full form (they were opening small forms with only that specific field in them) from validation issues.

**FEATURE:** Allow redirecting to forms after closing simple modals

When calling `store.dispatch('openModal', config)` you may now specify a `config.redirectTo` object with `{ uri, path, field }` which will trigger a specific form to open when the modal closes. The `uri` and `path` are for the form, and the `field` is the specific field that should be focused. This is great for pickers and other attached buttons in complex-list items, as you can see below:

Dev / Modder Notes:

Individual `inputs` now get an `initialFocus` prop (along with `name`, `data`, `schema`, and `args`) which will be the path for the field that should be focused. If your input can be focused (text, buttons, etc), please do a quick check on `$mount` against the `name` prop to see if your `input` should focus itself. Because inputs are varied (and may be complicated), this is left to individual inputs to deal with.

* [simple example - text](https://github.com/clay/clay-kiln/blob/cab0c2fa1eb3d7b557c2ec9e323967a2338433b3/inputs/text.vue#L136)
* [using 3rd party libraries - codemirror](https://github.com/clay/clay-kiln/blob/cab0c2fa1eb3d7b557c2ec9e323967a2338433b3/inputs/codemirror.vue#L182)
* [complex logic - wysiwyg](https://github.com/clay/clay-kiln/blob/cab0c2fa1eb3d7b557c2ec9e323967a2338433b3/inputs/wysiwyg.vue#L741)
* [passing prop to child component - simple-list](https://github.com/clay/clay-kiln/blob/cab0c2fa1eb3d7b557c2ec9e323967a2338433b3/inputs/simple-list.vue#L112)

### DIFF
--- a/lib/toolbar/actions.js
+++ b/lib/toolbar/actions.js
@@ -61,7 +61,6 @@ export function closeModal({ commit, state, dispatch }) {
       // component exists and is in the body (not a head component)
       dispatch('select', componentEl);
     }
-    console.log('focus back on the form:', uri, path, field);
     return dispatch('focus', { uri, path, initialFocus: field, el });
   }
 }
--- a/lib/toolbar/actions.js
+++ b/lib/toolbar/actions.js
@@ -12,6 +12,7 @@ import {
   SHOW_SNACKBAR,
   HIDE_SNACKBAR
 } from './mutationTypes';
+import { getFieldEl, getComponentEl } from '../utils/component-elements';
 
 /**
  * @module toolbar
@@ -42,8 +43,27 @@ export function openModal({ commit, dispatch }, config) {
   return dispatch('unfocus').then(() => commit(OPEN_MODAL, config));
 }
 
-export function closeModal({ commit }) {
+export function closeModal({ commit, state, dispatch }) {
+  const redirectTo = _.get(state, 'ui.currentModal.redirectTo');
+
   commit(CLOSE_MODAL);
+
+  // if `redirectTo` was configured when opening the modal, re-open a specific form
+  // after closing this modal
+  if (redirectTo) {
+    const uri = redirectTo.uri,
+      path = redirectTo.path, // field or group to open the form at
+      field = redirectTo.field, // field to focus on after opening the form
+      el = getFieldEl(uri, path),
+      componentEl = el && getComponentEl(el);
+
+    if (componentEl) {
+      // component exists and is in the body (not a head component)
+      dispatch('select', componentEl);
+    }
+    console.log('focus back on the form:', uri, path, field);
+    return dispatch('focus', { uri, path, initialFocus: field, el });
+  }
 }
 
 // open and close confirmation modals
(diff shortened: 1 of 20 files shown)


lib/toolbar/actions.js
```diff
@@ -42,8 +43,27 @@ export function openModal({ commit, dispatch }, config) {
...
+      path = redirectTo.path, // field or group to open the form at
+      field = redirectTo.field, // field to focus on after opening the form
+      el = getFieldEl(uri, path),
+      componentEl = el && getComponentEl(el);
+
+    if (componentEl) {
+      // component exists and is in the body (not a head component)
+      dispatch('select', componentEl);
+    }
+    console.log('focus back on the form:', uri, path, field);
```
byronhulcher: Probably don't need this

PR #1298, opened 2018-08-16
fix zeros not saving on text inputs

### LINKED ISSUE #1280: When opening product component number input clears zero value
Please write up one bug report OR one idea per GitHub issue. For example, if you have multiple (different) ideas, write up multiple GitHub issues.

## Bug
when a number input is given a value of zero in kiln, the input box clears when edit dialog is opened

rmfarrell:
`parseFloat` is causing the issue here. It's being called on an empty string. Keen handles zeros correctly without the `NaN` conversion. I could investigate that further but 🤷‍♂️ 

closes #1280

### DIFF
--- a/inputs/text.vue
+++ b/inputs/text.vue
@@ -110,7 +110,7 @@
       // every time the value of the input changes, update the store
       update(val) {
         if (this.isNumerical) {
-          val = parseFloat(val, 10);
+          val = parseFloat(val === '' ? 0 : val);
         } else if (_.isString(val)) {
           // remove 'line separator' and 'paragraph separator' characters from text inputs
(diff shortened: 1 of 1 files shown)


inputs/text.vue
```diff
@@ -110,7 +110,8 @@
       // every time the value of the input changes, update the store
       update(val) {
         if (this.isNumerical) {
-          val = parseFloat(val, 10);
```
rmfarrell: I 🔪 'd this `10` because I don't think `parseFloat` takes a 2nd arg(?)

nelsonpecora: yuuup, that's `parseInt`

PR #1312, opened 2018-08-29
add deep focus support

### LINKED ISSUE #1295: allow focus decorator to link into nested fields
The focus decorator (`data-editable`) should be able to link into nested fields, i.e. fields inside `complex-list` items.

Example

```
<ul>
  {{#each items }}
    <li data-editable="items.{{@index}}.title">{{ title }}</li>
  {{/each}}
</ul>
```

nelsonpecora:
fixes #1295

Allows `data-editable` to point to deeply-nested fields (i.e. fields inside of `complex-list` inputs)

### DIFF
--- a/lib/core-data/components.js
+++ b/lib/core-data/components.js
@@ -49,7 +49,14 @@ function isCamelCase(str) {
  * @throws {Error} if not camelCased
  */
 function assertCamelCasedProps(item, name) {
-  if (_.isString(item) && !isCamelCase(item)) {
+  if (_.isString(item) && _.includes(item, '.')) {
+    // deep path! make sure all parts of it are camelcased
+    let nonCamelCasedPart = _.find(item.split('.'), (part) => !isCamelCase(part));
+
+    if (nonCamelCasedPart) {
+      log.error(`Properties in component data must be camelCased: ${name} → ${item} (try "${_.camelCase(nonCamelCasedPart)}")`);
+    }
+  } else if (_.isString(item) && !isCamelCase(item)) {
     log.error(`Properties in component data must be camelCased: ${name} → ${item} (try "${_.camelCase(item)}")`);
   } else if (_.isObject(item)) {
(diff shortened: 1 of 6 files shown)


rmfarrell:
I think we need to support array syntax here. There is no way  developer will use a dot syntax to find an item in an array.

lib/core-data/components.js
```diff
@@ -49,7 +49,14 @@ function isCamelCase(str) {
  * @throws {Error} if not camelCased
  */
 function assertCamelCasedProps(item, name) {
-  if (_.isString(item) && !isCamelCase(item)) {
+  if (_.isString(item) && _.includes(item, '.')) {
+    // deep path! make sure all parts of it are camelcased
+    let nonCamelCasedPart = _.find(item.split('.'), (part) => !isCamelCase(part));
```
rmfarrell: think this can be `const`

lib/forms/overlay.vue
```diff
@@ -132,7 +132,7 @@
         </ui-tabs>
         <div v-else class="input-container-wrapper" :style="{ 'max-height': `calc(100vh - ${formTop} - 56px)`}">
           <div class="input-container">
-            <field v-for="(field, fieldIndex) in sections[0].fields" :key="fieldIndex" :name="field" :data="fields[field]" :schema="schema[field]" :initialFocus="initialFocus"></field>
+            <field v-for="(field, fieldIndex) in sections[0].fields" :key="fieldIndex" :name="field" :data="fields[field]" :schema="schema[field] || getFieldSchema(field)" :initialFocus="initialFocus"></field>
```
rmfarrell: I weakly prefer move the `schema[field]` here to the `getFieldSchema()` method just because I don't like in-template programming.

nelsonpecora: if we call `getFieldSchema()`, logic will run on every render. This way the `schema[field]` logic is actually only run when `schema` or `field` change

PR #1315, opened 2018-08-29
Fix numerical collapse

rmfarrell:
Also logs error if wrong type of input is used.

### DIFF
--- a/inputs/text.vue
+++ b/inputs/text.vue
@@ -54,6 +54,10 @@
   import label from '../lib/utils/label';
   import UiTextbox from 'keen/UiTextbox';
   import attachedButton from './attached-button.vue';
+  import logger from '../lib/utils/log';
+
+  const validInputTypes = ['text', 'search', 'url', 'tel', 'password', 'multi-line'],
+    log = logger(__filename);
 
   export default {
     props: ['name', 'data', 'schema', 'args', 'initialFocus'],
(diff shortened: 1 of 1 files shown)


rmfarrell:
screen shot 2018-08-29 at 5 51 55 pm

inputs/text.vue
```diff
@@ -64,11 +68,8 @@
     },
     computed: {
       type() {
-        if (this.args.type === 'multi-line' || !this.args.type) {
-          return 'text';
-        } else {
-          return this.args.type;
-        }
+        return  this.args.type === 'multi-line' || !this.args.type ?
```
nelsonpecora: what is this line even

rmfarrell: ah, I'm just turning this if into a ternary. eslint is stripping the `()` which I'd ordinarily keep in so it doesn't look confusing.

inputs/text.vue
```diff
@@ -130,17 +135,20 @@
       },
       enableInput() {
         this.isDisabled = false;
-      }
+      },
```
nelsonpecora: ,,,,,,,,,,

inputs/text.vue
```diff
@@ -130,17 +135,20 @@
...
       if (this.initialFocus === this.name) {
-        const offset = _.get(this, '$store.state.ui.currentForm.initialOffset');
 
         this.$nextTick(() => {
-          if (_.includes(['text', 'search', 'url', 'tel', 'password', 'multi-line'], this.type)) {
-            // selection range is only permitted on text-like input types
-            setCaret(this.$el, offset, this.data);
+  
+          // validate input type
+          if (!validInputTypes.includes(this.type)) {
```
nelsonpecora: good

PR #1323, opened 2018-09-04
Amphora v7

zonika:
This is a breaking change.

Covers:
- Page/Layout metadata ("state") operations
- layout is now its own Object in the store
- various updates to support the above
- using clayutils when possible
- random bug fixes


template.handlebars
```diff
@@ -28,6 +28,10 @@
         {{#each @root._componentSchemas}}
           window.kiln.preloadSchemas['{{ this.name }}'] = {{{ stringify (yaml (read this.schema)) }}};
         {{/each}}
+        {{#if @root._layoutData}}
+          window.kiln.layout = {{{ stringify @root._layoutData }}};
```
nelsonpecora: `layout` or `layout.data`?

lib/core-data/components.js
```diff
@@ -78,10 +79,18 @@ function assertCamelCasedProps(item, name) {
...
+  let schema, result, name, prefix;
 
-  let schema, result;
+  if (isComponent(uri)) {
+    name = getComponentName(uri);
+    prefix = `state.schemas['${name}']`;
+  } else if (isLayout(uri)) {
+    name = getLayoutName(uri);
+    prefix = 'state.layout.schema';
+  } else {
```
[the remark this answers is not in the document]
zonika: actually, looking at the code again, it looks like this function supports being called with a _component name_ or a uri, so going to keep the third branch just in case that functionality is used anywhere

template.handlebars
nelsonpecora: ahh ok

PR #1324, opened 2018-09-05
collapsible component lists

nelsonpecora:
* add `collapse: true` to `_componentList` config to enable
* will always display a placeholder with the placeholder `text`
* allows hiding/showing of list children

Empty List:

list_empty

Open List:

list_open

Collapsed List:

list_collapsed


lib/decorators/placeholder.vue
```diff
@@ -236,6 +280,15 @@
...
     methods: {
+      toggleCollapse() {
+        if (this.isCollapsed) {
+          this.isCollapsed = false;
+          this.$el.parentNode.classList.remove('kiln-collapsed');
+        } else {
+          this.isCollapsed = true;
+          this.$el.parentNode.classList.add('kiln-collapsed');
+        }
+      },
```
rmfarrell: Would this work here?
```js
this.isCollapsed = !this.isCollapsed
// ...add/remove conditional here or classList.toggle()
```
More importantly, is there a more Vue-friendly way to do this than external DOM manipulation? Seems fragile

docs/manipulating-components.md
```diff
@@ -46,6 +46,30 @@ myList:
 
 ![](images/list_placeholder.png)
 
+#### Collapsible Lists
+
+Basic lists may also be collapsed. This is useful if your components aren't a 1:1 WYSIWYG match with their public-facing view, for example if you have components that only display in `edit` mode.
```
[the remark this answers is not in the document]
nelsonpecora: I like it!

lib/decorators/placeholder.vue
```diff
@@ -66,6 +78,13 @@
     }
   }
 
+  // collapsible placeholders
+  .collapsible-component-list > .kiln-placeholder {
+    height: 68px !important;
+    margin: 0 0 20px;
+    min-height: 68px !important;
```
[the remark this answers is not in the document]
nelsonpecora: I should add a comment here. Basically this is overriding the normal (inline style attr) height stuff.

lib/decorators/placeholder.vue
```diff
@@ -135,7 +154,14 @@
 
 <template>
   <div :class="placeholderClass" :style="{ minHeight: placeholderHeight }" :ref="uid">
-    <ui-button v-if="isComponent" :disabled="!isActive" class="placeholder-add-component" icon="add" :color="placeholderButtonColor" @click.stop.prevent="openAddComponentPane">{{ addComponentText }}</ui-button>
+    <div v-if="isComponent && isCollapsible" class="placeholder-label-collapsible">
+      <span class="placeholder-text">{{ text }}</span>
+      <div class="ui-button-group">
+        <ui-button v-if="isEmptyList" :disabled="!isActive" class="placeholder-add-component" icon="add" :color="placeholderButtonColor" @click.stop.prevent="openAddComponentPane">{{ addComponentText }}</ui-button>
+        <ui-icon-button v-else class="placeholder-collapse-button" :icon="collapseIcon" :tooltip="collapseTooltip" :color="placeholderButtonColor" @click.stop.prevent="toggleCollapse"></ui-icon-button>
+      </div>
```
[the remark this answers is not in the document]
nelsonpecora: Ah yeah, I can remove it. Originally I was going to show both, but we should show the Add Component button only when it's empty. 👍

lib/decorators/placeholder.vue
```diff
@@ -135,7 +154,14 @@
...
+    </div>
+    <ui-button v-else-if="isComponent" :disabled="!isActive" class="placeholder-add-component" icon="add" :color="placeholderButtonColor" @click.stop.prevent="openAddComponentPane">{{ addComponentText }}</ui-button>
```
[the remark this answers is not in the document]
nelsonpecora: Hmm, I don't think having the three conditions written out with `v-if`s is terrible, but I'm going to add comments to them (which will also be helpful if we end up destructuring this code in the future if placeholders become too complex)

lib/decorators/placeholder.vue
nelsonpecora: Haha yeah I could just do that 🙃 (I forgot about `classList.toggle()`)

As for the fragility, I'll add a check to make it less brittle but there isn't really a more vue-y way to do this (since it's reaching out to a non-vue thing)

rmfarrell: it looks like you're passing the parentHeight to `$options`. Would it be useful to instead pass the whole parent reference? Esp if the height ever changes it would be better to track that within Vue anyway.

nelsonpecora: The parent reference might change on re-render

rmfarrell: oh dang. and the Vue component can't be notified?

lib/decorators/placeholder.vue
```diff
@@ -164,7 +195,9 @@
 
   export default {
     data() {
-      return {};
+      return {
+        isCollapsed: false
```
rmfarrell: isn't this true initially?

nelsonpecora: I was thinking about that, and I think I'd be fine with having it collapsed initially.

rmfarrell: u mean expanded? !isCollapsed  = expanded, no?

PR #1327, opened 2018-09-17
Convert null/undefined values to empty string

rmfarrell:


inputs/text.vue
```diff
@@ -24,7 +24,7 @@
 <template>
   <ui-textbox
     :autosize="false"
-    :value="String(data)"
+    :value="(data === null || data === undefined) ? '' : String(data)"
```
nelsonpecora: `typeof data === 'undefined'`?

PR #1336, opened 2018-09-25
add curly rule, fix handful of issues

nelsonpecora:
I noticed we neglected the `curly` rule, which can (and, in my experience, does) lead to bugs and reduces code clarity.


nelsonpecora:
merging because the coverage change isn't meaningful for this PR

PR #1339, opened 2018-10-01
Update mode/client/upgrade docs

rmfarrell:
- Remove legacy content (~controller.js~)
- More clarity around model
- Explain what a client.js file should look like
- Break each concern into its own subsection


fxisco:
This issue is been addressed here: https://github.com/clay/clay-kiln/pull/1399, with new documentation platform

PR #1347, opened 2018-10-12
range input

### LINKED ISSUE #905: idea: slider / range behavior
This could be easily implemented with https://github.com/NightCatSama/vue-slider-component

nelsonpecora:
* closes #905
* allows single-point and dual-point ranges
* uses material design spec
* shows tooltips when focusing handles (if enabled)
* full keyboard support

**Note: Screenshots are not real world use cases**

Single point, no tooltips, no custom min/max labels

Single point, tooltips, custom min/max labels

Dual point, tooltips, custom max label

### DIFF
--- a/inputs/range.vue
+++ b/inputs/range.vue
@@ -66,7 +66,7 @@
       height: 100%;
     	position: relative;
       width: 100%;
-    	z-index: 1;
+    	z-index: 2;
     }
 
     .noUi-connects {
@@ -188,7 +188,17 @@
       position: absolute;
     	white-space: nowrap;
     	text-align: center;
-      transform: translate(-50%, 50%);
+    }
+
+    // the first value comes after the first marker,
+    // so it's the second div
+    .noUi-value:nth-of-type(2) {
+      transform: translate(0, 50%);
+    }
+
+    // the last value is always the last div
+    .noUi-value:last-of-type {
+      transform: translate(-100%, 50%);
     }
 
     .noUi-marker {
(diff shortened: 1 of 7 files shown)


rmfarrell:
This is really lovely. 

I'd just say that in general, you could kill a lot of complexity if these were two separate components: `range` and `slider`

inputs/range.vue
```diff
@@ -0,0 +1,411 @@
...
+              if (val === max) {
+                return maxLabel;
+              }
+            }
+          }
+        }
+      });
+
+      // add keyboard support
+      _.each(findAll(el, '.noUi-handle'), (handle, index) => {
```
rmfarrell: consider native JS here

inputs/range.vue
```diff
@@ -0,0 +1,411 @@
...
+      slider.create(el, {
+        start: this.start,
+        step: step,
+        range: { min, max },
+        format: {
+          to(val) {
+            return step >= 1 ? parseInt(val) : val;
+          },
+          from(val) {
+            return step >= 1 ? parseInt(val) : val;
```
rmfarrell: Can you get away with a simple typecast here? `Number(val)`

inputs/range.vue
nelsonpecora: NodeLists aren't arrays, so IE in particular doesn't have a native `forEach` method on their prototype 😄

nelsonpecora:
I was considering having separate `slider` and `range` inputs, but that'd lead to a bunch of duplicated functionality and styles, and the terms themselves are interchangeable (the actual differentiation is how many points a slider/range has, which is not necessarily limited to 2, conceptually).

inputs/range.vue
rmfarrell: that's true for IE but you can call the prototype in the contexts of a NodeList, if we're using the Babel polyfill it adds this to NodeList and we don't support IE anyway.

rmfarrell:
I'm still a 👍 for this, I just think we should consider mixins in the future. the shared methods can make use of the spread operator and you wouldn't need all these `if` statements. Having one vs. multiple sliders is where a lot of the logic tends to diverge.

inputs/range.vue
nelsonpecora: fixed!

inputs/range.vue
zonika: I've found that `Array.from(NodeList).forEach` works natively

package.json
```diff
@@ -160,5 +161,7 @@
   "peerDependencies": {
     "amphora-search": "^6.2.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "nouislider": "^12.0.0"
```
zonika: this makes me sad!!!!!!!!

nelsonpecora: wait why

nelsonpecora: (nouislider is a really good, pretty lightweight library that implements a lot of functionality that is just completely unsupported with native html sliders)

nelsonpecora: oh it doesn't need to be a regular dep, right. good catch!!!

zonika: i liked having 0 deps!!!!!!!

Issue #1282, opened 2018-07-16
Bug: Empty page areas are treated as layout-level


fraddski:
When editing an empty page area, it is treated as though it exists at the layout level and must be edited using layout mode. Once at least one component is added, the area can be edited in page mode as expected.

PR #1359, opened 2018-12-12
fix computed placeholder active-ness on empty page-level component lists

zonika:
fixes https://github.com/clay/clay-kiln/issues/1282


lib/decorators/placeholder.vue
```diff
@@ -193,6 +194,12 @@
     return get(options.uri, options.path).schema;
   }
 
+  function isPageArea(uri, path) {
```
jonwinton: Function comment

Issue #1363, opened 2018-12-14
Don't persist properties added by handlebars templates to component data


zonika:
Right now, variables created by handlebars templates are being added to component data that's persisted by kiln. Ideally, these variables should only live within the scope of the handlebars template and the component data stored by kiln's shared store should not include these values.

PR #1365, opened 2018-12-14
preload component data that hasn't been mutated by handlebars

zonika:
...if it's available from amphora-html. See https://github.com/clay/amphora-html/pull/56

part of the fix for https://github.com/clay/clay-kiln/issues/1363


lib/preloader/actions.js
```diff
@@ -113,7 +113,7 @@ export default function preload({ commit, dispatch }) {
     site = locals && normalizeSiteData(locals.site),
     layoutSchema = _.get(window, 'kiln.layout.schema'),
     tabulaRasa = _.assign({}, defaultState),
-    components = _.reduce(data, reduceComponents, {}),
+    components = _.reduce(data._componentData || data, reduceComponents, {}),
```
zonika: baccwards compatibility

Issue #1351, opened 2018-11-19
BUG: If a custom validator throws an uncaught Error, no custom validators will report errors


byronhulcher:
If I have custom validator A that returns an array with an error object, and validator B, which throws an uncaught exception (for instance: `TypeError: Cannot read property 'thing' of undefined`), then the error from validator A will not be reported in the Health check, and a user would be able to successfully publish.

PR #1366, opened 2018-12-17
catch errors thrown by custom validators

zonika:
fixes https://github.com/clay/clay-kiln/issues/1351

### DIFF
--- a/lib/validators/actions.js
+++ b/lib/validators/actions.js
@@ -56,9 +56,7 @@ function runValidator(state) {
           };
         }
       })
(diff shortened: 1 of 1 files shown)


lib/validators/actions.js
```diff
@@ -52,6 +55,9 @@ function runValidator(state) {
             }, [])
           };
         }
+      })
+      .catch((e) => {
```
rmfarrell:
```suggestion
      .catch(log.error)
```

PR #1369, opened 2019-01-23
Custom global validators

### LINKED ISSUE #1368: Kiln Global Validation
## Idea
How can Kiln validators be run on global data? For example, how can Kiln validate a custom url?

## Proposed Solution/Implementation
We need an option for creating a global error with UI affordances for opening global panes/drawers.

fxisco:
## Description

This PR adds global validators to kiln. To enable this in the site we can include validators the same way we do for components: add a validator to `window.kiln.validators`. The main difference relies on the validator object, is needed to include `scope: metadata` to identify it as a global validator.

e.g.:

**Metadata validation for all pages**

include `scope: 'metadata'` in object and on type indicate if is `error` or `warning`

```js
'use strict';

module.exports = {
  label: 'Lorem Ipsum',
  description: 'Lorem ipsum dolor sit amet, consectetur...',
  type: 'error',
  scope: 'metadata',
  validate(meta) {
    if (meta.authors.length < 1) {
      return [{
        preview: 'Blah blah'
      }];
    }

    return [];
  }
};
```

**Metadata validation for specific page**

include `scope: 'metadata'` in object. **NOTE**: include `uri: <URI>,`

```js
'use strict';

module.exports = {
  label: 'Lorem ipsum',
  description: 'Lorem ipsum dolor sit amet, consectetur...',
  type: 'error',
  scope: 'metadata',
  uri: 'cjp2sl6c50000cyv2gt73xbwt',
  validate(meta) {
    if (meta.authors.length < 1) {
      return [{
        preview: 'Blah blah'
      }];
    }

    return [];
  }
};
```

## Approach taken

- Filter validations using condition scope , type and uri (for specific page).
- Running validators with a different function that fetch the metadata first and then use this info to validate all the metadata validators.

Note: we are able to create metadata warnings if necessary.

# Ticket

Closes #1368 

### WIP

- Missing tests
- Missing comments


fxisco:
@scottnash we should not add anything special. Here an example(https://github.com/nymag/sites/pull/7308/files)

if we don't set scope, it is consider a global validator, i.e. all the ones that already exist. To be able to make validations with the metadata we should specify `scope: 'metadata'` and what we are going to receive as argument in the validate functions will we the metadata and not the component state as normal.

PR #1382, opened 2019-02-25
Deep linking

scottnash:
https://app.zenhub.com/workspaces/clay-repos-596524da4a8a0769f2f03175/issues/clay/clay-kiln/1377

Opening nav/drawers now sets the url-hash so all drawers can be deep linked to, including plugins, if they are opened using the openDrawer or toggleDrawer action.  Any element can also set the url hash by calling the setHash action and passing it an url object, which can be read and reacted to from store.state.url to enable even deeper linking. 

These changes also combine the openNav and openDrawer functionality, and combines the data for the mobile and desktop nav buttons, so updating nav items no longer needs to be done in two places.

Feature Branch:
http://deep-linking-test.nymag.sites.aws.nymetro.com/

### DIFF
--- a/lib/deep-linking/actions.js
+++ b/lib/deep-linking/actions.js
@@ -12,14 +12,8 @@ import label from '../utils/label';
  * @module deep-linking
  */
 
-function openHashedForm(store, urlProps) {
-  let isAdmin, component, instance, path, foundComponent, uri, el, group, groupEl, initialFocus;
-
-  // get user permissions
-  // note: as clay always opens pages in page edit mode (not layout),
-  // we'll have to switch modes if the deep link is to a layout component
-  // (and the user has permission to edit it)
-  isAdmin = _.get(store, 'state.user.auth') === 'admin';
+function checkHashedForm(store, urlProps) {
+  let component, instance, path, foundComponent, initialFocus;
 
   // If we've got `urlProps`, let's access them
   component = _.get(urlProps, 'component', null);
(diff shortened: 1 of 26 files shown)


lib/deep-linking/actions.js
```diff
@@ -30,11 +24,25 @@ function openHashedForm(store, urlProps) {
...
-  // Get the group from the schema with the `path`
-  group = get(uri, path);
+  if (!foundComponent) {
+    return null; // if the hash in the url doesn't result in a found component, return null to avoid an error being thrown
+  } else {
+    return openHashedForm(store, foundComponent, path, initialFocus);
+  }
+}
+
+function openHashedForm(store, foundComponent, path, initialFocus) {
```
fxisco: some function documentation might help too here

lib/drawers/actions.js
```diff
@@ -27,5 +21,30 @@ export function toggleDrawer(store, name) {
  * @param  {Object} store
  */
 export function closeDrawer(store) {
+  store.dispatch('clearHash');
```
fxisco: does the order matters?

lib/nav/nav-menu.vue
```diff
@@ -141,65 +147,117 @@
         return Object.keys(_.get(window, 'kiln.navButtons', {}));
       },
       displayNavMenu() {
-        return !!_.get(this.$store, 'state.ui.currentNav');
+        const navShouldBeOpen = this.menuOptions.findIndex( (option) => option.id === this.currentDrawer ) >= 0 || this.customButtons.find((button) => button === this.currentDrawer);
```
fxisco: You can abstract the condition as you did with `!this.navBackgroundIsOpen`

scottnash:
Feature Branch for testing added http://deep-linking-test.nymag.sites.aws.nymetro.com/

lib/drawers/drawer.vue
```diff
@@ -148,82 +150,115 @@
...
-          }];
-        }
+        const state = _.get(this.$store, 'state'),
+          errors = _.get(state, 'validation.errors', []),
+          warnings = _.get(state, 'validation.warnings', []),
+          hashTabName = _.get(state, 'url.sites'), // state.url.sites is just the second param in the url hash object
+          openHealth = errors.length > 0 || warnings.length > 0 || hashTabName === 'health';
+
+
+        let tabs = [
```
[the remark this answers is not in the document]
scottnash: I think that's a good idea and should be combined with the toolbarOptions in edit-toobar.vue perhaps, since those are the buttons that open the drawers that show these tabs.  Might be clearer if it was object something like:
[
  {
    drawer: {
    ...drawerProperties,
    tabs: [
      {
        ...tabProperties
      },
      {
        ...tabProperties
      }
    ]
  }
  },
    {
    drawer: {
      ...drawerProperties,
      tabs: [
        {
          ...tabProperties
        },
        {
          ...tabProperties
        }
      ]
    }
  }
]

That way if we ever need to add or remove a drawer, it could all be done in one place.

lib/drawers/actions.js
scottnash: Order doesn't matter here

lib/nav/nav-menu.vue
scottnash: done.

lib/drawers/drawer.vue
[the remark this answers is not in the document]
fxisco: This will add a lot of value to the project

PR #1390, opened 2019-03-13
Restore to previous json

scottnash:
https://app.zenhub.com/workspaces/clay-repos-596524da4a8a0769f2f03175/issues/clay/clay-kiln/1005

Adds Restore to Published Functionality to Pages on the Published tab.  Button is disabled when no un published changes or in draft state. User is required to enter their username and is warned that what they are doing can not be undone before proceeding. 

Can be see in action on http://restore-from-published.nymag.sites.aws.nymetro.com/?edit=true#

### DIFF
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -368,22 +368,23 @@ export function addCreatedComponentsToPageArea(store, { newComponents,  currentU
     const uri = component[refProp] || currentURI;
 
     delete component[refProp]; // remove uri from actual data, since we're not doing a cascading PUT
+
+
     return modelSave(uri, component)
       .then((savedData) => addToQueue(save, [uri, savedData, false], 'save'))
       .then((savedData) => publish(uri, savedData, null, false, store))
       .then((savedData) => modelRender(uri, savedData))
       .then((savedData) => store.commit(UPDATE_COMPONENT, {uri, data: savedData}))
       .then(() => uri);
-  }))
-    .then((uris) => {
-      if (replace) {
-        // replace current uri when creating new list
-        return _.slice(data, 0, index).concat(uris).concat(_.slice(data, index + 1));
-      } else {
-        // just add new components after the index
-        return _.slice(data, 0, index + 1).concat(uris).concat(_.slice(data, index + 1));
-      }
-    })
+  })).then((uris) => {
+    if (replace) {
+      // replace current uri when creating new list
+      return _.slice(data, 0, index).concat(uris).concat(_.slice(data, index + 1));
+    } else {
+      // just add new components after the index
+      return _.slice(data, 0, index + 1).concat(uris).concat(_.slice(data, index + 1));
+    }
+  })
     .then((newList) => {
       if (replace) {
         store.commit(REMOVE_COMPONENT, { uri: currentURI });
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -91,8 +91,6 @@ function clientSave(uri, data, oldData, store, {eventID, snapshot, paths}) { // 
     })
     .then((savedData) => {
       // kick api call off in the background
-      console.log(uri);
-      console.log(savedData);
       addToQueue(save, [uri, savedData, false], 'save') // add hooks=false to prevent models from re-running on server
         // note: we don't care about the data we got back from the api
         .catch((e) => {
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -91,6 +91,8 @@ function clientSave(uri, data, oldData, store, {eventID, snapshot, paths}) { // 
     })
     .then((savedData) => {
       // kick api call off in the background
+      console.log(uri);
+      console.log(savedData);
       addToQueue(save, [uri, savedData, false], 'save') // add hooks=false to prevent models from re-running on server
         // note: we don't care about the data we got back from the api
         .catch((e) => {
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -361,14 +361,26 @@ function addComponentsToPageArea(store, {currentURI, path, replace, components})
     });
 }
 
-export function addCreatedComponentsToPageArea(store, { newComponents,  currentURI, path, replace, index, data}) {
+/**
+ * add components to a page area when they have already been created
+ * when the page is being reverted to the published version for instance
+ * @param {object} store
+ * @param {array} newComponents
+ * @param {string} currentURI
+ * @param {string} path
+ * @param {boolean} replace
+ * @param {index} number
+ * @param {data} array
+ * @param {forceRender} boolean
+ * @returns {Promise}
+ */
+export function addCreatedComponentsToPageArea(store, { newComponents,  currentURI, path, replace, index, data, forceRender}) {
   // note: we have to explicitly save these (rather than relying on the cascading PUT of the parent),
   // because they exist in page data
   return Promise.all(_.map(newComponents, (component) => {
     const uri = component[refProp] || currentURI;
 
     delete component[refProp]; // remove uri from actual data, since we're not doing a cascading PUT
-
 
     return modelSave(uri, component)
       .then((savedData) => addToQueue(save, [uri, savedData, false], 'save'))
@@ -390,7 +402,7 @@ export function addCreatedComponentsToPageArea(store, { newComponents,  currentU
         store.commit(REMOVE_COMPONENT, { uri: currentURI });
       }
       // save page
-      return store.dispatch('savePage', { [path]: newList });
+      return store.dispatch('savePage', { [path]: newList, forceRender });
     });
 }
 
--- a/lib/component-data/actions.js
+++ b/lib/component-data/actions.js
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import { removeElement, find } from '@nymag/dom';
 import { isPage } from 'clayutils';
-import { UPDATE_COMPONENT, REVERT_COMPONENT, REMOVE_COMPONENT, CURRENTLY_SAVING, OPEN_ADD_COMPONENT, CLOSE_ADD_COMPONENT } from './mutationTypes';
+import { UPDATE_COMPONENT, REVERT_COMPONENT, REMOVE_COMPONENT, CURRENTLY_SAVING, OPEN_ADD_COMPONENT, CLOSE_ADD_COMPONENT, CURRENTLY_RESTORING } from './mutationTypes';
 import { getData, getSchema } from '../core-data/components';
 import { refProp, refAttr, layoutAttr, editAttr, componentListProp, componentProp, getComponentName } from '../utils/references';
 import { save as modelSave, render as modelRender } from './model';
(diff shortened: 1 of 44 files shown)


lib/drawers/publish-page.vue
```diff
@@ -220,6 +52,13 @@
       <span class="archive-help">You may archive any page that isn't published (or scheduled to be published). Archived pages will not show up in the Clay Menu unless you explicitly filter for them.</span>
       <ui-button class="archive-submit" buttonType="button" type="primary" color="red" :disabled="isScheduled || isPublished || isArchived" @click.stop="archivePage(true)">Archive</ui-button>
     </ui-collapsible>
+
+
+    <!-- restore from published -->
+    <ui-collapsible class="publish-section publish-archive" title="Restore Published Version">
+      <span class="archive-help">You may override ALL local changes to the page by Restoring the Page to the Published Version.  This can not be undone.</span>
```
jonwinton: How about something like:


> You may override all changes to this page by restoring to the most recently published version of the page by clicking the button below. **Warning:** this action cannot be undone, you will lose all changes applied to this post since the last publish.

I'm mainly worried the word `local` won't translate. Once a user makes a change in clay it's technically global since any user who comes in between edits will see what has already been done.

lib/drawers/publish-page.vue
```diff
@@ -220,6 +52,13 @@
       <span class="archive-help">You may archive any page that isn't published (or scheduled to be published). Archived pages will not show up in the Clay Menu unless you explicitly filter for them.</span>
       <ui-button class="archive-submit" buttonType="button" type="primary" color="red" :disabled="isScheduled || isPublished || isArchived" @click.stop="archivePage(true)">Archive</ui-button>
     </ui-collapsible>
+
+
+    <!-- restore from published -->
```
jonwinton: Random question: should we hide this tab until the page is published? Or maybe disable the dropdown if the page isn't published? I have no idea what's the best

lib/component-data/actions.js
```diff
@@ -357,38 +357,55 @@ function addComponentsToPageArea(store, {currentURI, path, replace, components})
...
+ * @param {array} newComponents
+ * @param {string} currentURI
+ * @param {string} path
+ * @param {boolean} replace
+ * @param {index} number
+ * @param {data} array
+ * @param {forceRender} boolean
+ * @returns {Promise}
+ */
+export function addCreatedComponentsToPageArea(store, { newComponents,  currentURI, path, replace, index, data, forceRender}) {
```
manuelurenah: Remember to add spacing when destructuring an object
```suggestion
export function addCreatedComponentsToPageArea(store, { newComponents,  currentURI, path, replace, index, data, forceRender }) {
```

lib/drawers/publish-page.vue
scottnash: I debated about that when building it, but just went with disabling the actual restore button when it's not published or has no changes. But hiding it completely would work as well because I guess, why show it at all if the user can't do anything with it.

lib/drawers/publish-page.vue
scottnash: I'll update the copy

lib/toolbar/type-restore-published-version-modal.vue
```diff
@@ -0,0 +1,63 @@
...
+      .ui-button {
+        margin-left: 16px;
+      }
+    }
+  }
+</style>
+
+<template>
+  <div class="ui-reason type-reason">
+    <div class="ui-reason__message"><p>Are you sure you want to restore to the published version? This action cannot be undone and <b>all of the changes since the page was last published will be lost permanently</b>.</p><p>Please type your email address to verify you wish to continue.</p></div>
```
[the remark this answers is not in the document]
scottnash: That works for me.  Wasn't really sure what to use for confirmation.  I'll change it to a phrase to be matched.

lib/component-data/actions.js
```diff
@@ -357,38 +357,55 @@ function addComponentsToPageArea(store, {currentURI, path, replace, components})
...
+      .then((savedData) => store.commit(UPDATE_COMPONENT, {uri, data: savedData}))
+      .then(() => uri);
+  })).then((uris) => {
+    if (replace) {
+      // replace current uri when creating new list
+      return _.slice(data, 0, index).concat(uris).concat(_.slice(data, index + 1));
+    } else {
+      // just add new components after the index
+      return _.slice(data, 0, index + 1).concat(uris).concat(_.slice(data, index + 1));
+    }
```
manuelurenah: To reduce a couple lines of codes, we could use ternary operator to store the value of the `index` (since that's the only value that changes from the statement) and have a single `return` statement. Here's a quick example
```js
const seletcedIndex = replace ? index : index + 1;

return _.slice(data, 0, selectedIndex).concat(uris).concat(_.slice(data, index + 1));
```

lib/component-data/mutations.js
```diff
@@ -35,6 +35,10 @@ export default {
     _.set(state, 'ui.currentlySaving', isSaving);
     return state;
   },
+  [CURRENTLY_RESTORING]: (state, isSaving) => {
+    _.set(state, 'ui.currentlyRestoring', isSaving);
```
manuelurenah: What do you think if we rename `isSaving` in here to be `isRestoring`?
```suggestion
    _.set(state, 'ui.currentlyRestoring', isRestoring);
```

lib/drawers/publish-page.vue
```diff
@@ -290,6 +132,10 @@
       currentTitle: (state) => state.page.state.title,
       isAdmin: (state) => state.user.auth === 'admin',
       isLayoutPublished: (state) => state.layout.state.published,
+      headComponents: (state) => state.page.data.head,
+      hasChanges: (state) => {
+        return hasPageChanges(state);
+      },
```
manuelurenah: We could have this in one line
```js
hasChanges: (state) => hasPageChanges(state),
```

lib/drawers/publish-page.vue
```diff
@@ -564,6 +410,112 @@
...
+            });
+          });
+        });
+      },
+      saveComponents(components, path, forceRender = false) {
+        // Save Components and Rerender in the page
+        return new Promise((resolve) => {
+          const arrComponents = [];
+
+          components.forEach((component) =>{
```
manuelurenah: Please mind spacing in here
```suggestion
          components.forEach((component) => {
```

lib/drawers/publish-page.vue
```diff
@@ -564,6 +410,112 @@
...
+      saveComponent(component) {
+        // Update the changed components in the Vue Store
+        // without this, vue won't be in sync with the rendered content
+        // and will revert to the unrestored content when editing
+        let newData = component,
+          uri = component['_ref'];
+
+        delete newData['_ref'];
+
+        newData = { ...this.$store.state.components[uri], ...newData};
```
manuelurenah: Please mind spacing
```suggestion
        newData = { ...this.$store.state.components[uri], ...newData };
```

lib/toolbar/edit-toolbar.vue
```diff
@@ -181,14 +181,7 @@
...
-        const pubTime = _.get(state, 'page.state.publishTime'), // latest published timestamp
-          upTime = _.get(state, 'page.state.updateTime'); // latest updated timestamp
-
-        if (pubTime && upTime) {
-          return isAfter(upTime, pubTime);
-        } else {
-          return false;
-        }
+        return hasPageChanges(state);
       },
```
manuelurenah: We can have this in a single line
```js
hasPageChanges: (state) => hasPageChanges(state),
```

lib/drawers/publish-page.vue
```diff
@@ -564,6 +410,112 @@
...
+          });
+        });
+      },
+      saveComponents(components, path, forceRender = false) {
+        // Save Components and Rerender in the page
+        return new Promise((resolve) => {
+          const arrComponents = [];
+
+          components.forEach((component) =>{
+            const data = JSON.parse(JSON.stringify(component).replace(new RegExp('@published','g'),''));
```
manuelurenah: This is clever!!! But are there any drawbacks about having to convert an object to string, then back to object?

lib/utils/history.js
```diff
@@ -72,3 +74,17 @@ export function getLastEditUser(itemState, currentUser) {
...
+    pubTime = _.get(pageState, 'publishTime'), // latest published timestamp
+    upTime = _.get(pageState, 'updateTime'), // latest updated timestamp
+    mostRecentPageHistory = pageState.history[pageState.history.length - 1],
+    lastHistoryAction = mostRecentPageHistory && mostRecentPageHistory.action ? mostRecentPageHistory.action : '';
+
+  if (pubTime && upTime && lastHistoryAction !== 'restore') {
+    return isAfter(upTime, pubTime);
+  } else {
+    return false;
+  }
```
manuelurenah: We could have this as ternary operator
```js
return pubTime && upTime && lastHistoryAction !== 'restore'
  ? isAfter(upTime, pubTime)
  : false;
```

lib/drawers/publish-page.vue
jonwinton: Honestly this isn't anything I feel strongly about so I defer to your judgement!

scottnash: I went ahead and hid it when page isn't published or there are no changes.  It matches the unpublish button behavior this way, since that button is also not there until published.

lib/component-data/actions.js
scottnash: 👍

lib/component-data/actions.js
scottnash: I switched to the ternary, although this wasn't actually code that I changed.  It was already like that.  I just broke one function into two, so it looks as though I added it.

lib/component-data/mutations.js
scottnash: Definitely should be isRestoring, I just cut and pasted the CURRENTLY_SAVING function above it and forgot to update param name.

lib/drawers/publish-page.vue
scottnash: 👍

lib/drawers/publish-page.vue
scottnash: 👍

lib/drawers/publish-page.vue
scottnash: 👍

lib/toolbar/edit-toolbar.vue
scottnash: 👍

lib/drawers/publish-page.vue
scottnash: No drawbacks that I can think of, and it's faster than looping through the nested object looking for properties with the @published url and updating.

lib/utils/history.js
scottnash: 👍

lib/drawers/publish-page.vue
manuelurenah: Nice! Loved it

PR #1391, opened 2019-03-25
Add docusaurus basic structure

oscarpolanco:
Adding basic docs structure to the repo.

# Steps for testing the site locally:
- Go to the `website` folder
- Install all dependencies using `npm install`
- Start your local server using `npm run start`
- Test locally at [localhost](http://localhost:3000/)

# Steps to test the build:
- Go to the `website` folder
- Install all dependencies using `npm install`
- Start your local server using `npm run build`
- Check that now you got a `build` folder under the `website` directory

# Steps to test the doc versioning:
- Go to the `website` folder
- Install all dependencies using `npm install`
- Start your local server using `npm run version 0.0.0` (number of your version)
- Start your local server using `npm run start`
- Go to [localhost](http://localhost:3000/)
- On the navbar you should see the number of your new doc version
- Go to [version page](http://localhost:3000/clay-kiln/versions)
- Check if your version is on the list

# First deploy process (manual):
- Create a local branch call `gh-pages` from `master`
- Go to the `website` folder
- Install packages using `npm install`
- Do a build using `npm build`
- go back to the `root` folder
- Add the files and commit using this command => `git add -f website/build/Kiln/ && git commit -m"deploy files"`
- Push the static files using this command => `git  subtree push --prefix website/build/Kiln origin gh-pages`
## Note:
- This Process I will do it when the` PR` is approved
- We only need the static files on the `gh-pages` branch and the commit history only will show docs update and important remember that the static files are part of the `.gitignore` that's why we add and push this way.
- Any suggestions to have a better first deploy process are more than welcome 😄 

# General Notes: 
- There are some duplicate files until we got the live docs page
- We still don't have the final design
- Leave some of the default files for future references

PR #1392, opened 2019-03-27
Add workflow file for the deploy

oscarpolanco:
Add deploy process for the docs using the [docusaurus action repo](https://github.com/clay/docusaurus-github-action/) and the search of the page

# Steps for testing the site locally:
- Install [act](https://github.com/nektos/act#installation)
- Go to the [`.github` ](https://github.com/clay/clay-kiln/tree/add_action_search/.github) folder
- On the `main.workflow` update the[`branch`](https://github.com/clay/clay-kiln/blob/add_action_search/.github/main.workflow?short_path=a334a5b#L8) line to use `branch action_deploy`
- Use the [`act`](https://github.com/nektos/act#commands) command to simulate a `push` event. The build should be `successful`
- On the `main.workflow` update the[` branch`](https://github.com/clay/clay-kiln/blob/add_action_search/.github/main.workflow?short_path=a334a5b#L8) line to use `branch master`
- Use the [`act`](https://github.com/nektos/act#commands) command to simulate a `push` event.  The build should be `decline`

Note:
- ~We need to merge this [PR](https://github.com/clay/clay-kiln/pull/1391) first~
- ~We need a deploy key to complete this task~

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -76,7 +76,7 @@ const siteConfig = {
 
   // Open Graph and Twitter card images.
   ogImage: '',
-  twitterImage: '',
+  twitterImage: ''
 
   // Show documentation's last contributor's name.
   // enableUpdateBy: true,
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,0 +1,15 @@
+workflow "Deploy to GitHub Pages" {
+  on = "push"
+  resolves = ["Build and push docs"]
+}
+
+  action "Filter branch" {
+  uses = "actions/bin/filter@master"
+  args = "branch master"
+}
+
+  action "Build and push docs" {
+  needs = ["Filter branch"]
+  uses = "clay/docusaurus-github-action@master"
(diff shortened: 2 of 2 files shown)


website/siteConfig.js
```diff
@@ -77,6 +77,11 @@ const siteConfig = {
   // Open Graph and Twitter card images.
   ogImage: '',
   twitterImage: '',
+  algolia: {
+    apiKey: process.env.ALGOLIA_API_KEY,
+    indexName: 'TBD',
```
fxisco: what is the use of this `indexName`? When we need to change it?

oscarpolanco: This is the name of the Algolia index where your records will be pushed. We don't determine this; they send us this name with the key.

oscarpolanco: Here a little more of the config file
https://community.algolia.com/docsearch/config-file.html#index_name
And the `Docusaurus` setup docs
https://docusaurus.io/docs/en/search

PR #1399, opened 2019-04-17
Add model, update and client content on the component docs

oscarpolanco:
Add the content regarding this PR => https://github.com/clay/clay-kiln/pull/1339

Need to discuss if we need to add the `client.js` and `update.js` on the `kiln` docs

# Steps for testing the site locally:
- Go to the `website` folder
- Install all dependencies using `npm install`
- Start your local server using `npm run start`
- Test locally at [localhost master](http://localhost:3000/clay-kiln/docs/component#modeljs)
- Compare with the current docs [localhost](https://docs.clayplatform.com/clay-kiln/docs/component#model-and-controller)

### DIFF
--- a/docs/component.md
+++ b/docs/component.md
@@ -35,27 +35,54 @@ The root element of `<body>` components may include other attributes, such as `d
 
 ---
 
-## Model and Controller
+## Model.js
+`models.js` is an optional file that a component directory may contain. Its role is to expose two lifecycle hooks that allow an author to transform the component's data: save, which executes before the model is saved and render, which executes before the HTML is rendered.
 
-Models \(`model.js`\) and controllers \(`client.js` or `controller.js`\) are two types of JavaScript files your components may contain \(a third — `upgrade.js` — is [only used in Amphora](http://clay.github.io/amphora/docs/upgrade.html)\). Models allow components to perform logic on their data before saving and rendering, while controllers add client-side functionality to components. Both are optional.
+`save` and `render` are both optional, but `model.js` must export at least one of these functions. Both hooks take the same signature and are expected to return the data. For asynchronous operations, `save` and `render` may also return a promise.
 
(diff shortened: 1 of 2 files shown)


docs/component.md
```diff
@@ -35,28 +36,55 @@ The root element of `<body>` components may include other attributes, such as `d
 
 ---
 
-## Model and Controller
-
-Models \(`model.js`\) and controllers \(`client.js` or `controller.js`\) are two types of JavaScript files your components may contain \(a third — `upgrade.js` — is [only used in Amphora](http://clay.github.io/amphora/docs/upgrade.html)\). Models allow components to perform logic on their data before saving and rendering, while controllers add client-side functionality to components. Both are optional.
+## Model.js
+`models.js` is an optional file that a component directory may contain. Its role is to expose two lifecycle hooks that allow an author to transform the component's data: save, which executes before the model is saved. and render, which executes before the HTML is rendered.
```
fxisco: remove period after word `saved`

docs/component.md
```diff
@@ -35,28 +36,55 @@ The root element of `<body>` components may include other attributes, such as `d
...
 };
 ```
 
-> If in doubt, use the `save()` method, as it only runs when data is saved. Calling the `render()` method \(every time a component renders\) to do slow or intensive logic will slow down your component for end-users.
+> If you're unsure of which method to use `save()` is usually preferable, as it only runs when data is saved. Calling the `render()` method \(every time a component renders\) to do slow or intensive logic will slow down your component for end-users.
 
 Models must be compiled and added to `window.kiln.componentModels` for Kiln to call them when saving and re-rendering components.
 
+## client.js
+client.js is the component's code that is exectued in the client in view mode only (i.e., not edit mode). This file is optional. It expects a single default export which takes the following signature, where el is root HTML element of the component:
```
fxisco: typo `exectued `

docs/component.md
oscarpolanco: Nice catch

PR #1401, opened 2019-04-29
snackbSnackbar properties

scottnash:
Exposing more properties on the snackbar ui element (the small alert bar that appears at the bottom left of the page when the user completes actions such as publishing/unpublishing a page, adding a user, etc.) to make it more configurable, particularly by plugins, who may wish to change the location of the bar.

When showing the snackbar the following properties can now be set:
* ***duration*** - time in milliseconds the snackbar is shown in this container. Defaults to 5000
* ***queueSnackbars*** - Whether or not snackbars should be queued and shown one after the other. False by default, creating a new snackbar while one is visible will cause the visible one to immediately transition out for the new one. Set to true to ensure that each snackbar is shown for its complete duration.
* ***position*** - The position of snackbars relative to the container. One of 'left', 'center' or 'right'.  Defaults to 'left'
* ***transition*** - The show/hide transition of snackbars in the container. One of 'slide' or 'fade'. Defaults to 'slide'


lib/toolbar/view-toolbar.vue
```diff
@@ -139,3 +105,42 @@
...
+    }
+
+    .ui-snackbar-container {
+      @include confirm-layer();
+
+      bottom: 0;
+      position: fixed;
+    }
+  }
+</style>
```
manuelurenah: Why did we move the styles to the bottom of the file? 😛

lib/toolbar/edit-toolbar.vue
```diff
@@ -337,9 +343,8 @@
     }),
     watch: {
       snackbar(val) {
-        if (val) {
+        if (Object.keys(val).length > 0) {
```
manuelurenah: We could have the condition without evaluating `> 0`
```suggestion
        if (Object.keys(val).length) {
```

lib/toolbar/view-toolbar.vue
scottnash: Because that's the standard Vue order of tags. Template, Script, then Styles.

lib/toolbar/edit-toolbar.vue
scottnash: 👍 changed

lib/toolbar/view-toolbar.vue
scottnash: From the [vuejs.org style guide](https://vuejs.org/v2/style-guide/) 
"Single-file components should always order `<script>`, `<template>`, and `<style>` tags consistently, with `<style>` last, because at least one of the other two is always necessary."

PR #1402, opened 2019-04-29
Snackbar properties

scottnash:
Updating docusaurus docs for snackbar


scottnash:
Just a documentation update

PR #1425, opened 2019-05-31
Add the project name condition on the docs site config

oscarpolanco:
Use the process `env` to add the site on the config.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -6,7 +6,7 @@
  */
 
 const repoUrl = 'https://github.com/clay/clay-kiln';
-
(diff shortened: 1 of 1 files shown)


website/siteConfig.js
```diff
@@ -6,7 +6,7 @@
  */
 
 const repoUrl = 'https://github.com/clay/clay-kiln';
-
+projectName = process.env.PROJECT_NAME || 'site';
```
manuelurenah: Shouldn't the default be `clay-kiln`?

oscarpolanco: Nice catch

PR #1429, opened 2019-06-04
Update the docs and website  README

oscarpolanco:
Update the `README` of the `docs` and `website` directory with the info and step to work with the documentation

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,31 @@
+# Clay Documentation
+
+On this directory, we have the pre-release documentation written using markdown. This file has the `.md` extension and as we mentioned before, these are pre-release files if you don't add a documentation version yet.
+
+## Markdown options
+
+These files have the following structure at the beginning of the file:
+
+```
+---
+id: idOftheFile
+title: Title of your file
+sidebar_label: Title of the sidebar
+---
+
+My new content here..
+```
+
+The previous example has a couple of the most important part to identify the content that you wish to add to your documentation. Here is a more detailed explanation  on the [Markdown feature](https://docusaurus.io/docs/en/doc-markdown) that you can use.
+
+## Test documentation page
+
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -0,0 +1,31 @@
...
+---
+id: idOftheFile
+title: Title of your file
+sidebar_label: Title of the sidebar
+---
+
+My new content here..
+```
+
+The previews example has a couple of the most important part to identify the content that you wish to add to your documentation. Here is a more detail explanation  on the [Markdown feature](https://docusaurus.io/docs/en/doc-markdown) that you can use.
```
fxisco:
```suggestion
The previews example has a couple of the most important part to identify the content that you wish to add to your documentation. Here is a more detail explanation on the [Markdown feature](https://docusaurus.io/docs/en/doc-markdown) that you can use.
```

website/README.md
```diff
@@ -56,30 +60,60 @@ root-directory
...
+yarn examples versions
 ```
+
+or
+
+```sh
+npm run examples versions
+```
+
+Then you proceed to create a version of the docs that you need in this case need to match the version of your current `package.json`. To do this we will use the following command followed the number of the version that you want.
```
fxisco:
```suggestion
Then you proceed to create a version of the docs that you need in this case needed to match the version of your current `package.json`. To do this we will use the following command followed the number of the version that you want.
```

website/README.md
```diff
@@ -56,30 +60,60 @@ root-directory
...
 ```
+
 or
-```
+
+```sh
 $ npm version X.X.X
 ```
 
+This will create all the files need it to handle the version that you just build but this will be necessary only in 2 cases:
```
fxisco:
```suggestion
This will create all the files needed to handle the version that you just build but this will be necessary only in 2 cases:
```

oscarpolanco:
All feedback has been handled!
Thanks, @manuelurenah @fxisco

PR #1442, opened 2019-06-13
Kilnjs validation documentation update

scottnash:
Adding the ability for kilnjs to use custom validators and add the results to the health tab.  Documentation detailing this is updated in the kilnjs.md file in the docs.


lib/drawers/health.vue
```diff
@@ -156,6 +165,16 @@
...
+    <div v-for="(warning, index) in kilnjsWarnings" class="publish-warning" :key="`publish-kilnjs-warning-item-${index}`">
+      <span class="warning-label">{{ warning.label }}</span>
+      <span class="warning-description">{{ warning.description }}</span>
+      <span class="validation-info">Go To Components</span>
+      <ul class="validation-items">
+        <li v-for="(item, idx) in warning.items" class="validation-item" :key="`publish-warning-validation-item-${idx}`">
+          <span class="validation-item-location" :class="{ 'validation-item-link': item.uri && item.field }" @click.stop="openLocation(item.uri, item.field, item.path, item.location)">{{ item.location }}</span> <span v-if="item.preview" class="validation-item-preview">{{ item.preview }}</span>
+        </li>
+      </ul>
+    </div>
```
manuelurenah: The error (lines 148-157) and warning (lines 168-177) elements are basically the same, do you think it would be possible to create a reusable component to display these elements?

docs/kilnjs.md
```diff
@@ -66,9 +66,81 @@ You can attach events to the schema inputs. The events you can attach vary depen
...
+    }
+  ];
+```
+
+And the titleLength validator would look like this.
+```js
+const { getPreviewText }= window.kiln.utils.validationHelpers;
+
+export const label = 'Title Length',
+  description = 'Title Length Can\'t be longer 40 characters when using an H2',
```
PedroAlbR:
```suggestion
  description = 'Title Length can\'t be longer 40 characters when using an H2',
```

docs/kilnjs.md
```diff
@@ -66,9 +66,81 @@ You can attach events to the schema inputs. The events you can attach vary depen
...
+  type = 'error';
+
+export function kilnjsValidate([title, size]) {
+  return  size === 'h2' && title.length > 40
```
PedroAlbR:
```suggestion
  return size === 'h2' && title.length > 40
```

docs/kilnjs.md
```diff
@@ -84,6 +156,27 @@ module.exports = (schema) => {
...
+You have access to all of the custom validators loaded in the kiln.validators object, so if you wish to write a reusable validation function that can be used across components, you could add a function to any of the existing validators or add a new one. For example you could use the same custom penelopecruz validator inline rather than adding it to the schema.validation object. However, running the validator this way will not add the error/warning to the health tab and you will need to communicate the error to the user manually using the snackbar component.
+
+```js
+const validators = schema.fieldName.validators(),
+  error = validators.penelopeCruz.kilnjsValidate(schema.title.value());
+```
+
+#### Form Validation
+
+Along with input level validation, you can also perform form validation using kiln.js.  To do this, you add a function to the schame named `formValidation`. This function is called just before a form is saved and if it returns `false` then the form will not be saved.  The snackBar element can be used to convey a message to the user explaining why the form wasn't saved.
```
PedroAlbR:
```suggestion
Along with input level validation, you can also perform form validation using kiln.js.  To do this, you add a function to the schema named `formValidation`. This function is called just before a form is saved and if it returns `false` then the form will not be saved.  The snackBar element can be used to convey a message to the user explaining why the form wasn't saved.
```

docs/kilnjs.md
```diff
@@ -84,6 +156,27 @@ module.exports = (schema) => {
...
+
+```js
+  schema.formValidation = () => {
```
PedroAlbR: I think we should ident this block since its isolated code
```suggestion
schema.formValidation = () => {
  if (schema.size.value() === 'h2' && schema.title.value().length > 40) {
    kilnInput.showSnackBar({ message: 'When size is H2, the Title can\'t be longer than 40 chars', position: 'center' })
    return false;
  }

  return true;
};
```

lib/validators/actions.js
```diff
@@ -120,6 +120,86 @@ export function runMetaValidators(uri, errorValidators, warningValidators) {
...
+          validators[validationRule.validator].kilnjsValidate;
+}
+
+/**
+ * Run the kilnjsValidators
+ * @param {object} schemas & components
+ * @param {string} errorLevel
+ * @return {array}
+ */
+export function runKilnjsValidators({schemas = {}, components}, errorLevel) {
```
manuelurenah: Please mind spacing
```suggestion
export function runKilnjsValidators({ schemas = {}, components }, errorLevel) {
```

docs/kilnjs.md
```diff
@@ -66,9 +66,81 @@ You can attach events to the schema inputs. The events you can attach vary depen
...
+You can combine the power of custom validators with kiln.js by adding a validation property to the schema that specifies which field should be validated using which custom validator.
+```js
+  schema.validation = [
+    {
+      input: 'title', // the name of the field to be validated
+      validator: 'penelopeCruz' // the name of the custom validator to validate the field with
+    }
+  ];
+```
+In the custom validator file that lives in the services/kiln/validators folder, you construct a validator file in much the same way you would create one to be used by the schema file, only instead of exporting a function named validate, you export one called kilnjsValidate that validates on the value you pass it.
```
fxisco: Could we highlight `kilnjsValidate` ?

docs/kilnjs.md
```diff
@@ -66,9 +66,81 @@ You can attach events to the schema inputs. The events you can attach vary depen
...
+    index = text.toLowerCase().indexOf('penelope cruz');
+
+  return containsPenelopeCruz(value)
+    ? { label, description, type, preview: getPreviewText(text, index, 13) }
+    : false;
+}
+```
+Errors or warnings generated by kilnjsValidate appear on the health tab rather than giving immediate feedback, in the same way that ordinary custom validators work.
+
+You can create custom validators that validate based on the value of multiple fields as well as single fields.  If validation rules depend upon multiple fields, you pass an array of fields in the input property.  Just remember that the order of field names in the array is important because the custom validator function will recieve an array with the value of those fields as arguments in the same order as they appear in the original array.
```
fxisco: Maybe adding Warning signs to the remember phrase or showing it as an alert could add some value

lib/validators/actions.js
```diff
@@ -120,6 +120,86 @@ export function runMetaValidators(uri, errorValidators, warningValidators) {
...
+  const { validators } = window.kiln,
+    errors = [];
+
+  Object.keys(schemas).forEach((schemaName) => {
+    // Loop through all schemas on the page and check if they have a validation property
```
fxisco: In my understanding this can be implied from the code.

lib/validators/actions.js
```diff
@@ -120,6 +120,86 @@ export function runMetaValidators(uri, errorValidators, warningValidators) {
...
+    if (schemas[schemaName].validation ) {
+      // if they have a validation property, then loop through all the rules in the validation property array
```
fxisco: In my understanding this can be implied from the code.

lib/validators/actions.js
```diff
@@ -120,6 +120,86 @@ export function runMetaValidators(uri, errorValidators, warningValidators) {
...
+ * Run the kilnjsValidators
+ * @param {object} schemas & components
+ * @param {string} errorLevel
+ * @return {array}
+ */
+export function runKilnjsValidators({schemas = {}, components}, errorLevel) {
+  const { validators } = window.kiln,
+    errors = [];
+
+  Object.keys(schemas).forEach((schemaName) => {
```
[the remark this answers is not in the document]
scottnash: I broke this up into different functions and added tests

docs/kilnjs.md
scottnash: 👍

docs/kilnjs.md
scottnash: 👍

docs/kilnjs.md
scottnash: 👍

docs/kilnjs.md
scottnash: 👍

lib/drawers/health.vue
```diff
@@ -145,7 +145,16 @@
         </li>
       </ul>
     </div>
-
+        <div v-for="(error, index) in kilnjsErrors" class="publish-error" :key="`publish-kilnjs-error-item-${index}`">
```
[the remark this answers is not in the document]
scottnash: 👍

lib/drawers/health.vue
scottnash: Not only possible, but also a great idea.  Done.

lib/validators/actions.js
scottnash: 👍

docs/kilnjs.md
scottnash: 👍

docs/kilnjs.md
scottnash: 👍

lib/validators/actions.js
scottnash: Agree. I ended up rewriting this function and breaking it up into smaller ones, removing most of the comments as I did it.

lib/validators/actions.js
scottnash: Agree. I ended up rewriting this function and breaking it up into smaller ones, removing most of the comments as I did it.

lib/validators/actions.js
fxisco: uffff looks much better now! Nice! 👍

lib/forms/overlay.vue
```diff
@@ -21,15 +21,15 @@
           <ui-tab v-for="(section, index) in sections" :key="index" :title="section.title" :selected="initialSection === index">
             <div class="input-container-wrapper" :style="{ 'max-height': `calc(100vh - ${formTop} - 104px)`}">
               <div class="input-container">
-                <field v-for="(field, fieldIndex) in section.fields" :key="fieldIndex" :name="field.name" :data="fields[field.name]" :visibility="!field.schema || field.schema.visibility" :schema="field.schema || schema[field.name] || getFieldSchema(field.name)" :initialFocus="initialFocus"></field>
+                <field v-for="(field, fieldIndex) in section.fields" :key="JSON.stringify(field.schema) + fieldIndex" :name="field.name" :data="fields[field.name]" :visibility="!field.schema || field.schema.visibility" :schema="field.schema || schema[field.name] || getFieldSchema(field.name)" :initialFocus="initialFocus"></field>
```
manuelurenah: I think we can use template literals in here to concat the strings
```suggestion
                <field v-for="(field, fieldIndex) in section.fields" :key="`${JSON.stringify(field.schema)}${fieldIndex}`" :name="field.name" :data="fields[field.name]" :visibility="!field.schema || field.schema.visibility" :schema="field.schema || schema[field.name] || getFieldSchema(field.name)" :initialFocus="initialFocus"></field>
```

lib/validators/actions.js
```diff
@@ -120,6 +120,114 @@ export function runMetaValidators(uri, errorValidators, warningValidators) {
...
+
+/**
+ * Get the schemas that have a validation property
+ * @param {object} schemas
+ * @returns {object}
+ */
+export function getSchemasWithValidationRules(schemas = {}) {
+  const validationSchemas = {};
+
+  Object.keys(schemas).forEach((schemaName) => schemas[schemaName].validation ? validationSchemas[schemaName] = _.cloneDeep(schemas[schemaName]) : null );
```
manuelurenah: Can you please break this line into multiple lines? Just to make it more readable
```suggestion
  Object.keys(schemas).forEach((schemaName) => (
    schemas[schemaName].validation
      ? validationSchemas[schemaName] = _.cloneDeep(schemas[schemaName])
      : null
    )
  );
```

PR #1443, opened 2019-06-18
Kilnjs custom fetch

scottnash:
Adding a custom fetch function to kilnjs that adds caching and timeout functionality to the native fetch() function


docs/kilnjs.md
```diff
@@ -230,6 +230,34 @@ module.exports = (schema) => {
 
 `KilnInput` also provides its own set of custom methods.
 
+### fetch
+
+* ***fetch(url, options, cache, timeOutIn)*** - a function that wraps a standard fetch call that provides caching and throws a catchable error if the call takes longer than a set amount of time.
+  * url - The url of the endpoint to which you're making the fetch call
+  * options - A JSON object that holds the request options of a fetch call. Can be used to se the methods, headers, body, etc.  The same as you would use in a standard fetch().
```
manuelurenah:
```suggestion
  * options - A JSON object that holds the request options of a fetch call. Can be used to set the methods, headers, body, etc.  The same as you would use in a standard fetch().
```

docs/kilnjs.md
```diff
@@ -230,6 +230,34 @@ module.exports = (schema) => {
...
+  * cache - A boolean value that defaults to true. When true, it will cache future calls to the exact same url during this session. Set to false and it will make a new fetch call everytime it is invoked.
+  * timeOutIn - The number of milliseconds before the fetch will timeout.  Defaults to 5000. When the call timesout, the fetch is aborted and an error is thrown with the message 'Timeout'.
```
manuelurenah:
```suggestion
  * timeOutIn - The number of milliseconds before the fetch will timeout.  Defaults to 5000. When the call times out, the fetch is aborted and an error is thrown with the message 'Timeout'.
```

oscarpolanco:
```suggestion
  * timeOutIn - The number of milliseconds before the fetch will timeout.  Defaults to 5000. When the call times out, the fetch is aborted and an error is thrown with the message 'Timeout'.
```

docs/kilnjs.md
scottnash: 👍

docs/kilnjs.md
scottnash: 👍

docs/kilnjs.md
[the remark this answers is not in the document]
scottnash: "set", not "see", but definitely a typo.

docs/kilnjs.md
scottnash: 👍

PR #1447, opened 2019-06-21
Pubsub bug

scottnash:
wrapping savecomponent() call in a $nextTick to prevent race condition


lib/component-data/pubsub.js
```diff
@@ -4,6 +4,9 @@ import { pubProp, subProp, componentRoute, refProp, getComponentName } from '../
 import { ADD_SCHEMA } from './mutationTypes';
 import { PRELOAD_SUCCESS } from '../preloader/mutationTypes';
 import { props } from '../utils/promises';
+import Vue from 'vue';
+
+const vue = new Vue;
```
manuelurenah: If I'm not mistaken, I think we have to use `Vue()` in order to instantiate it
```suggestion
const vue = new Vue();
```

lib/component-data/pubsub.js
```diff
@@ -287,15 +290,20 @@ function updateComponentInstances({ eventID, snapshot, pubData, store }) {
       const instances = _.filter(Object.keys(store.state.components), (uri) => _.includes(uri, `${componentRoute}${name}/`) && !_.isEmpty(store.state.components[uri]));
 
       return Promise.all(_.compact(_.map(instances, (uri) => {
+        // the saveComponent dispatch is wrapped in a $nextTick to prevent a race condition where one component is subscribed to multiple values, which can result in each subsequent save overwriting the values in the previous ones
         if (scope && subscriberInScope({ scope, publisherURI, subscriberURI: uri, pubData, store })) {
           // subscriber is in scope!
-          return store.dispatch('saveComponent', { uri, data, eventID, snapshot });
+          return vue.$nextTick().then(()=> {
```
manuelurenah:
```suggestion
          return vue.$nextTick().then(() => {
```

lib/component-data/pubsub.js
```diff
@@ -287,15 +290,20 @@ function updateComponentInstances({ eventID, snapshot, pubData, store }) {
...
+            return store.dispatch('saveComponent', { uri, data, eventID, snapshot });
+          });
         } else if (scope) {
           // not in scope!
           return;
         } else {
           // no scope!
-          return store.dispatch('saveComponent', { uri, data, eventID, snapshot });
+          return vue.$nextTick().then(()=> {
```
manuelurenah:
```suggestion
          return vue.$nextTick().then(() => {
```

cruzanmo: A good example of what a style guide/linter would help with

jjpaulino:
```suggestion
          return vue.$nextTick().then(() => store.dispatch('saveComponent', { uri, data, eventID, snapshot }));
```

lib/component-data/pubsub.js
jjpaulino:
```suggestion
          return vue.$nextTick().then(() => store.dispatch('saveComponent', { uri, data, eventID, snapshot }));
```

lib/component-data/pubsub.js
scottnash: JavaScript lets you omit parens when declaring a new Object when there are no arguments, but I went ahead and added them to avoid confusion.

lib/component-data/pubsub.js
scottnash: 👍

lib/component-data/pubsub.js
scottnash: 👍

scottnash: 👍

lib/component-data/pubsub.js
scottnash: 👍

PR #1454, opened 2019-07-02
Add protocolfix

scottnash:


inputs/kiln-input.js
```diff
@@ -268,4 +268,16 @@ export default class KilnInput {
       return store.state.ui.currentForm ? store.state.ui.currentForm.fields[this.inputName] : null;
     }
   }
+
+  addProtocol(url, location) {
```
fxisco: Missing function description

scottnash: good catch!  added

PR #1455, opened 2019-07-02
Newaddprotocl

scottnash:
replacing old add protocol with new


inputs/kiln-input.js
```diff
@@ -72,7 +72,9 @@ export default class KilnInput {
   * @return {Promise}
   */
   getComponentData(uri) {
-    const url = this.addProtocol(uri);
+    const url = addProtocol(uri);
+
+    console.log(url);
```
manuelurenah: Please remove this log

lib/utils/urls.js
```diff
@@ -15,18 +15,24 @@ export function addPort(uri, location) {
...
-export function addProtocol(uri, location) {
-  location = location || /* istanbul ignore next: can't stub window.location */ window.location;
-  const hasProtocol = uri.indexOf(location.protocol) === 0;
+* Check url string for protocol and if it doesn't have one, add the protocol of the current url
+* @param {string} url
+* @param {object} location - only passed in testing
+* @return {string}
+*/
+export function addProtocol(url, location) {
+  const pattern = /^((http|https):\/\/)/;
```
manuelurenah: This can be set as a global at the top of the file so we don't need to create this property every time the `addProtocol` function is called

lib/utils/urls.js
```diff
@@ -15,18 +15,24 @@ export function addPort(uri, location) {
...
+  if (!pattern.test(url)) {
+    location = location || /* istanbul ignore next: can't stub window.location */ window.location;
+
+    url = `${location.protocol}//${url}`;
+  }
+
+  return url;
 }
 
+
```
manuelurenah: Please remove this extra line

PR #1457, opened 2019-07-08
Fix value condition in kiln-input.js

fxisco:
## FIX

This PR will help setting `false` value with `kiln-input.js`

### DIFF
--- a/inputs/kiln-input.js
+++ b/inputs/kiln-input.js
@@ -259,7 +259,7 @@ export default class KilnInput {
   * @return {string|number|object|array}
   */
   value(val) {
-    if (val) {
+    if (val != null) {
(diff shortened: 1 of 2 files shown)


inputs/kiln-input.js
```diff
@@ -259,7 +259,7 @@ export default class KilnInput {
   * @return {string|number|object|array}
   */
   value(val) {
-    if (val) {
+    if (val != null) {
```
manuelurenah: Please use `!==` instead
```suggestion
    if (val !== null) {
```

fxisco: With this condition we can avoid `null` and `undefined` at the same time

PR #1472, opened 2019-08-15
Subsite support

jpope19:
Adds the ability to search for pages by subsite. Subsites appear as another site inside of the page/my-pages view

### DIFF
--- a/lib/nav/page-list.vue
+++ b/lib/nav/page-list.vue
@@ -54,14 +54,18 @@
    */
   function getInitialSites() {
     const configSites = _.get(this.$store, 'state.url.sites'),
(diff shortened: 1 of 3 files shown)


lib/nav/page-list.vue
```diff
@@ -259,10 +286,10 @@
         this.isPopoverOpen = false;
       },
       selectSite(slug) {
-        const site = _.find(this.sites, s => s.slug === slug);
+        const site = _.find(this.sites, s => s.subsiteSlug === slug || s.slug === slug);
```
fxisco: I think we can create a function because is the same condition in 2 places

jpope19: fixed

Issue #1474, opened 2019-08-21
Clicking Publish menu doesn't trigger save on actively edited component


phillyDesignr:
## Description

Hello, Is it the intent that focus from a component being actively edited is not removed when the Publishing menu is clicked?

We are seeing an issue where our editors have focus on a page component and have made edits however, they have not triggered a save byway of removing focus or hitting the Enter button. They are then trying to immediately publish by opening the Publishing menu which, seems to not trigger a blur and therefore doesn’t trigger the component’s save prior to the page publish which, causes their active edit to not persist on the page.

## Steps to Reproduce

1. create a new or edit a page with an editable component like a heading
2. change something in the heading
3. do not trigger a save by clicking outside the component or hitting enter
4. click the Publishing menu button
5. the blur and therefore save is NOT triggered on the component

## Expected Behavior

when a user clicks anywhere off the component the save should be triggered

## Screenshots

Francisco Payés @fxisco has a video from our zoom meeting

### Desktop

 - OS: MacOs 10.14.6
 - Browser Chrome 
 - Version Version 76.0.3809.100 (Official Build) (64-bit)

PR #1475, opened 2019-08-23
feat: set focus on toolbar buttons when they are clicked to unfocus a…

fxisco:
## Feature

This PR helps the toolbar button to gain focus when is clicked. This way any input that is being edited will loose focus and as result will be saved.

Issue: https://github.com/clay/clay-kiln/issues/1474

### DIFF
--- a/lib/toolbar/edit-toolbar.vue
+++ b/lib/toolbar/edit-toolbar.vue
@@ -107,6 +107,7 @@
 
         <!-- display individual buttons on larger screens (viewport >= 600px) -->
         <component
+          :ref="option.id"
           size="large"
           v-for="option in activeToolBarOptions"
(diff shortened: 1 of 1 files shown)


lib/toolbar/edit-toolbar.vue
```diff
@@ -405,7 +407,6 @@
           const activeNav = savedTab || 'all-pages';
 
           this.$store.dispatch('showNavBackground', true);
-
```
manuelurenah: Please leave an empty line before a `return` statement to make ir more readable

fxisco: ohhhh there was an empty space that my linter removed. Thank you man.

PR #1477, opened 2019-09-05
fix: Handle keys args in select input

MariaFdezJ:
# FIX

This PR:
- Allows keys args.
- Allows same keys for label and value (helps specially with `_lists` that only have `text` and `count`).

### DIFF
--- a/inputs/select.test.js
+++ b/inputs/select.test.js
@@ -1,0 +1,42 @@
+import Select from './select.vue';
+
+describe('select input', () => {
+  const filterBySite = jest.fn(),
+    options = {
+      propsData: {
+        name: 'foo',
+        data: null,
+        schema: {},
+        args: { options: [1, 2, 3] }
+      },
+      mocks: {
+        $store: {
+          state: { site: { slug: 'test' } }
+        }
+      }
+    };
+
+  let wrapper;
+
+  filterBySite.mockReturnValue([1, 2, 3]);
+
+  test('Should use default computed keys when no keys are sent in args', () => {
+    options.propsData.args.keys = {};
+    wrapper = shallowMount(Select, options);
+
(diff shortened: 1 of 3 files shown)


inputs/select.vue
```diff
@@ -132,22 +132,36 @@
...
-            return val.name === 'None' || val.name === 'Default';
+            return val[label] === 'None' || val[label] === 'Default';
           });
 
         let fullOptions = propOptions.concat(this.listOptions);
 
+        if (this.equalLabelValueKeys) {
+          fullOptions = fullOptions.map((option) => ({
+            [label]: option[label],
+            value: option[label]
```
manuelurenah: We can avoid having the if condition in here and just use `this.valueKey`
```suggestion
            [this.valueKey]: option[label]
```

inputs/select.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  let wrapper;
+
+  filterBySite.mockReturnValue([1, 2, 3]);
+
+  beforeEach(() => {
+    options.propsData.args.keys = {};
+  });
+
+  test('default computed keys when none is sent in args', () => {
```
PedroAlbR:
```suggestion
  test('Should use default computed keys when no keys are sent in args', () => {
```

inputs/select.test.js
```diff
@@ -0,0 +1,53 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
MariaFdezJ: This actually is not need, so removing. Thanks!

MariaFdezJ:
[[clay-kiln]](https://trello.com/c/UUX54knx/367-clay-kiln)

MariaFdezJ:
[Handle keys args in select input](https://trello.com/c/UUX54knx/367-handle-keys-args-in-select-input)

PR #1480, opened 2019-09-12
Changes setProp behavior

manuelurenah:
## Description

- Changes `kiln-input.js` file to support `setProp` being called directly in any `kiln.js` file

## Example

```js
'use strict';

const KilnInput = window.kiln.kilnInput;

module.exports = schema => {
  schema.imageUrl = new KilnInput(schema, 'imageUrl');
  schema.imageAlt = new KilnInput(schema, 'imageAlt');

  schema.imageAlt.setProp('disabled', true);
  schema.imageUrl.setProp('visibility', false);

  return schema;
};

```

### DIFF
--- a/inputs/kiln-input.js
+++ b/inputs/kiln-input.js
@@ -1,14 +1,14 @@
-import store from '../lib/core-data/store';
-import * as api from '../lib/core-data/api.js';
 import { replaceVersion } from 'clayutils';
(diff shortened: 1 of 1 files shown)


manuelurenah:
Heads up! Most of the changes in this PR are from the linter and code styling, but the main change is in the `setProp` and `hide` methods 🏄

PR #1481, opened 2019-09-17
Set disabled flags on inputs

rhark1:
Current disabled functionality is not propagated to the actual UI components.  This prevents the UI components from displaying the correct disabled UI.  This PR just passes the disabled flag to the Keen UI components.


inputs/segmented-button.vue
```diff
@@ -63,6 +64,8 @@
         const data = this.data,
           assetLocation = _.get(this.$store, 'state.site.assetHost') || _.get(this.$store, 'state.site.assetPath');
 
+        console.log('options', this.args.options);
```
jpope19: remove log plz

inputs/wysiwyg.vue
```diff
@@ -695,6 +695,10 @@
         }
       });
 
+      if (this.schema.disabled) {
```
MariaFdezJ: Why are we taking it from `schema` here, and not from `props` as the others?

inputs/simple-list-input.vue
```diff
@@ -55,7 +57,7 @@
   import autocomplete from './autocomplete.vue';
 
   export default {
-    props: ['items', 'allowRepeatedItems', 'autocomplete', 'currentItem', 'disabled', 'isInitialFocus', 'ignoreComma'],
+    props: ['items', 'allowRepeatedItems', 'autocomplete', 'currentItem', 'disabled', 'isInitialFocus', 'ignoreComma', 'disabled'],
```
MariaFdezJ:
```suggestion
    props: ['items', 'allowRepeatedItems', 'autocomplete', 'currentItem', 'disabled', 'isInitialFocus', 'ignoreComma'],
```

It's already added, after`currentItem` and is adding `classes` for that. Is it actually needed for this component?

inputs/range.vue
```diff
@@ -230,6 +230,10 @@
           el.noUiSlider.on(key, this.schema.events[key]);
         });
       }
+
+      if (this.disabled) {
+        el.setAttribute('disabled', this.disabled);
+      }
```
MariaFdezJ: Can't we add it directly to this line: https://github.com/clay/clay-kiln/blob/2d5759f9ca024859363d3562052a505ccb8f5475/inputs/range.vue#L40 as `:disabled="disabled"` to keep consistency with the others?

inputs/wysiwyg.vue
rhark1: I didn't want anyone to confuse it with the `isDisabled` data prop being used in the same component

inputs/range.vue
rhark1: The attribute is being set on the slider created in the `mounted` hook, not the actual element

PR #1483, opened 2019-10-01
Simple list: Update count when removing item

MariaFdezJ:
# Fix
When you add an item to the list, it will update the list with the new item and increment the count, but if you remove the item from the list, the count is not decremented.

### DIFF
--- a/inputs/simple-list.test.js
+++ b/inputs/simple-list.test.js
@@ -1,0 +1,42 @@
+import SimpleList from './simple-list.vue';
+
+describe('simple list input', () => {
+  const options = {
+      propsData: {
+        name: 'foo',
+        data: null,
+        schema: {},
+        args: {}
+      }
+    },
+    items = [
+      { text: 'test1', count: 1 },
+      { text: 'test2', count: 2 }
+    ];
+
+  let wrapper;
+
+  test('Should decrease count when there is a removedItem', () => {
+    wrapper = shallowMount(SimpleList, options);
+    wrapper.vm.removedItem = { text: 'test2' };
+
+    expect(wrapper.vm.handleRemoveItem(items)).toStrictEqual([
+      { text: 'test1', count: 1 },
+      { text: 'test2', count: 1 }
+    ]);
+  });
+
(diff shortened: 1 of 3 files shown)


MariaFdezJ:
[As a clay user, I want to update the count list when add or remove an Item so that the list has accurate data.](https://trello.com/c/4LhZcjjD/368-as-a-clay-user-i-want-to-update-the-count-list-when-add-or-remove-an-item-so-that-the-list-has-accurate-data)

inputs/simple-list.vue
```diff
@@ -233,14 +234,47 @@
...
+       * Handles remove item from list
+       * @param {Array.<Object>} items
+       */
+      handleRemoveItem(items = []) {
+        // validate that the list has items with these properties
+        const stringProperty = getProp(items, 'text'),
+          countProperty = getProp(items, 'count');
+
+        if (stringProperty && countProperty) {
+          const itemIndex = getItemIndex(items, (this.removedItem || {}).text, 'text');
```
PedroAlbR: If you set `this.removedItem` to an empty object when you delete it and by default, you will never have to do this validation

inputs/simple-list.test.js
```diff
@@ -0,0 +1,42 @@
+import SimpleList from './simple-list.vue';
+
+describe('simple list input', () => {
```
PedroAlbR:
```suggestion
describe('SimpleList', () => {
```

inputs/simple-list.test.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    items = [
+      { text: 'test1', count: 1 },
+      { text: 'test2', count: 2 }
+    ];
+
+  let wrapper;
+
+  test('Should decrease count when there is a removedItem', () => {
+    wrapper = shallowMount(SimpleList, options);
```
PedroAlbR: You could add this to a `beforeEach`

inputs/simple-list.vue
MariaFdezJ: I'm taking this suggestion, however, since we don't have control over the lists, I understand this is something worth checking.

PR #1491, opened 2019-12-27
Fix: Validation Bug

CristianUser:
## What does this PR do?
Solve a validation bug working with kiln.js files.

### How to reproduce the bug.
- Add kiln.js file without validation property to a component (in case you don't have any component with kiln.js file).
- Create a new article and let it empty.
Now If to publish the validators don't execute. And appear this in the console:
Screen Shot 2019-12-26 at 11 06 07 AM

### Manual testing steps?
Do the same that I mentioned before and have to work correctly.

### DIFF
--- a/lib/validators/actions.js
+++ b/lib/validators/actions.js
@@ -147,8 +147,8 @@ export function runKilnjsValidators(schemas = {}, components = {}) {
   Object.keys(components).forEach((uri) => {
     const componentName = getComponentName(uri);
 
-    if (_.get(schemas, `${componentName}.validation`)) {
-      _.get(schemas, `${componentName}.validation`, []).forEach((validationRule) => {
+    if (schemas[componentName]) {
+      schemas[componentName].validation.forEach((validationRule) => {
         if (validKilnjsValidator(validators, validationRule)) {
           const valueField = getKilnJsFieldAndValue(validationRule, components[uri]),
             validationError = validators[validationRule.validator].kilnjsValidate(valueField.value);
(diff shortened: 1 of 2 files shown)


lib/validators/actions.js
```diff
@@ -180,7 +180,7 @@ export function runKilnjsValidators(schemas = {}, components = {}) {
 export function getSchemasWithValidationRules(schemas = {}) {
   const validationSchemas = {};
 
-  Object.keys(schemas).forEach(schemaName => schemas[schemaName].validation ? validationSchemas[schemaName] = _.cloneDeep(schemas[schemaName]) : null);
+  Object.keys(schemas).forEach(schemaName => _.get(schemas, `${schemaName}.validation`, []) ? validationSchemas[schemaName] = _.cloneDeep(schemas[schemaName]) : null);
```
PedroAlbR: This validation is not needed since we are getting the schema name from from the schema, so `schemas[schemaName]` should always return something

lib/validators/actions.js
```diff
@@ -147,8 +147,8 @@ export function runKilnjsValidators(schemas = {}, components = {}) {
   Object.keys(components).forEach((uri) => {
     const componentName = getComponentName(uri);
 
-    if (schemas[componentName]) {
-      schemas[componentName].validation.forEach((validationRule) => {
+    if (_.get(schemas, `${componentName}.validation`)) {
```
PedroAlbR: We want to check if the component is there first, I think this doesn't need to change

CristianUser: You are right.

lib/validators/actions.js
CristianUser: But if the schema is not returned in the kiln.js file this throws an error and breaks the validators.

PedroAlbR: Makes sense. Maybe we should ignore that `kiln.js` or send a descriptive error.

PR #1505, opened 2020-11-19
Adds link validation for clay-paragraph, clay-subheader

mattoberle:
## Updated PR Description

This pull request implements link validation for QuillJS `<a>` tags embeded in the `clay-paragraph` and `clay-subheader` components. Validation is performed by overriding the default QuillJS `Link` class with a custom `SafeLink` sub class.

`SafeLink` sets `class="kiln-link-invalid"` on the `<a>` tag whenever a link does not meet any of the following criteria:

* The link is an anchor link (eg. `#section-2`).
* The link is relative to the site root (eg. `/section`).
* The link is a `mailto` link (eg. `mailto:test@example.com`).
* The link is in the format `{protocol}://{domain}/*`.

Kiln will block publishing when a page contains invalid links.
Invalid links are styled in **red** to make them easily distinguishable.

In addition to flagging invalid links, this pull request also _removes_ links applied to single punctuation and white-space characters. QuillJS requires users to select the full hyperlinked text to remove a link; this makes it very easy to leave a space or punctuation character hyperlinked. Tracking down those single-character links is like looking for a needle in a haystack.

#### Old PR Description

~This pull request contains several iterations of link validation.~

~The first commit demonstrates how we can subclass `formats/link` to create a Quill Blot that self-removes when invalid. This approach works but felt like a hack, it was ultimately scrapped.~

~The current state of the PR demonstrates how we can utilize `lib/validators/built-in/invalid-links.js` to surface validation errors to end users (while retaining their original input). The validators approach closely mirrors what we do in `lib/validators/edit-links.js` and feels most at-home in Kiln.~

~I have some reservations about the way `edit-links.js` and `invalid-links.js` work.
The fact that both validation functions are using regular expressions to pull out anchor tag values from partial HTML blobs feels fragile. It also feels redundant to iterate over the state twice to validate links.~

~**Note:**~

~Tests executed with `jest` pass on this branch, although we seem to have accumulated quite a few files that don't pass Kiln's `eslint` rules. The linting is what fails in CircleCI but I don't want to muddy this PR with a bunch of scattered style updates.~ This is fixed.

~**Note 2:**~

~Tests do not pass with the Node8 container runtime, but do pass with Node10+.
I've opened #1507 which will adjust the runtime used for tests (to reflect the currently supported Node versions).~ This is merged.

## Future Suggestions

These suggestions are likely **outside of the scope** of this pull request but would result in improvements to Kiln:

**One-iteration Validators**

It seems wasteful to have each built-in validator iterate over the entire state.
A more efficient pattern would be to compose all of the validators together and iterate over the state once.

**Behaviors**

@reubenson pointed out that Kiln has a concept of [behaviors](https://github.com/clay/clay-kiln/blob/efe67eed4ff81f505f9ff109c1f27738583422d7/lib/core-data/behaviors2input.js) that may allow components to opt-in/out of validation.


lib/validators/built-in/invalid-links.js
```diff
@@ -0,0 +1,60 @@
...
+ * Collects all invalid links from the page state and returns
+ * an array of validation errors.
+ *
+ * @param {Object} state
+ * @returns {Object[]} An array of validation errors.
+ */
+export function validate(state) {
+  const errors = [];
+
+  forEachComponent(state, (componentData, uri) => forEachField(state, componentData, uri, (field) => {
```
reubenson, line 41: this ends up combing over a lot of components and their properties. i wonder if it'd be worth whitelisting this only for generic clay components, e.g. `clay-paragraph` and `clay-subheader`. could also see an argument for this as is though, as a general all-purpose link validator

mattoberle: I was just checking out `forEachComponent` and noticed that it accepts a filter.
I think limiting it to `clay-paragraph` could be a good idea (although `edit-links.js` does not).

Maybe an initial release w/ `clay-paragraph` as a filter (to reduce the blast area) and then that can be loosened if the case presents itself.

mattoberle: I applied the filter for `clay-paragraph` and `clay-subheader`.

reubenson: fair enough. not a necessity for this PR, but there also things called [behaviors in kiln](https://github.com/clay/clay-kiln/blob/efe67eed4ff81f505f9ff109c1f27738583422d7/lib/core-data/behaviors2input.js) and maybe the best multi-purpose solution would to have link validation settable on the component schema level

reubenson (approved):

lib/validators/built-in/invalid-links.js
mattoberle: That's really good to know, I'm going to throw that into the PR description just to note.

james-owen (approved):

PR #1507, opened 2020-12-07
upgrades node-sass, runs tests on 10, 12, and 14 LTS

mattoberle:
This commit upgrades node-sass to work with newer node runtimes.
It also updates the CircleCI test routine to test all supported LTS versions (build and test).

The change should ensure that Kiln can be built on any supported runtime.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,27 +1,95 @@
+---
 version: 2
+
+references:
+  filter_all: &filter_all
+    filters:
+      branches:
+        only: /.*/
+      tags:
+        only: /.*/
+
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,27 +1,95 @@
...
+
 general:
   artifacts:
     - coverage
+
 jobs:
-  build:
+  test_node10:
     docker:
       - image: node:8.10.0
```
james-owen, line 33: should this be `node:10`?

mattoberle: Yes :man_facepalming: .

PR #1511, opened 2020-12-16
Use patch for list updates

jpope19:


lib/lists/actions.js
```diff
@@ -22,9 +22,27 @@ export function updateList(store, { listName, fn }) {
   store.commit(LIST_LOAD_PENDING, { listName });
 
   return getListFromAPI(prefix, listName)
+    // TODO check this error and make sure its a 404 before returning an empty array
     .catch(() => []) // allow adding to a new list
     .then(listItems => fn(_.cloneDeep(listItems)))
     .then(listItems => saveList(prefix, listName, listItems))
```
olsonpm, line 28: so it looks like the remaining lists using 'updateList' are new-pages and bookmarks, neither of which will ever reach a large number of items.

However if we keep updateList in the codebase, then
1. future clay devs may introduce it to update a large list without realizing
2. plugins may introduce bugs by dispatching an action for a large list.

so I could see value in forcing patchList to be used moving forward (i.e. remove updateList).  But no strong opinions there

olsonpm (approved):

lib/lists/actions.js
jpope19: yea i was hesitant to remove it from the codebase in case there were already existing plugins that used it. and because of this, felt like it'd be a waste of time re-writing the bookmarks/new-pages logic.

i'll definitely rework those two functions to use patchList though, if we think it'd be beneficial

olsonpm: ah yeah, the trouble of moving people off updateList when it is likely fine for 99% of lists - is probably not worth it.

so yeah i'm good with leaving this as is

mariamera (approved):

lib/lists/actions.js
```diff
@@ -16,15 +16,52 @@ export function getList(store, listName) {
...
+      } else {
+        throw error;
+      }
+    });
+}
+
+export function updateList(store, { listName, fn }) {
+  const prefix = _.get(store, 'state.site.prefix');
+
+  return getOrStartList(prefix, listName)
```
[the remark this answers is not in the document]
jpope19: good catch! this function is only used for bookmarks and new pages so it passed testing.

MariaFdezJ (approved):

PR #1532, opened 2021-08-10
add disable input support for ui-select

### LINKED ISSUE #1531: "locked" select input is not disabled
## Description

When the "lock" attached button is added to an input, it should initially be disabled, and only become enabled when the lock is clicked. Clicking the lock again should again disable the input. For the "select" input, it is not disabled.

## Steps to Reproduce

1. add `lock: true` to a field in your schema that has `input: select`
2. refresh the form
3. verify the lock button appears, and the select input is *not* disabled

## Expected Behavior

The select input should be disabled

## Screenshots

note that the input is enabled even when the lock is locked:
<video src="https://user-images.githubusercontent.com/1163/128856243-20aed984-538b-4a04-bfeb-7112166ee99e.mp4"></video>

macgyver:
Looks like this was implemented in a half-baked way - the `disabled` prop was being respected but the `isDisabled` value in `data` was being ignored. It has been updated to respect both values, with the prop taking precedence over the component state - I believe this is the intended behavior.

<video src="https://user-images.githubusercontent.com/1163/128879569-ce81100c-1eb9-4d4c-821d-93c9c56529ff.mp4
"></video>

closes #1531

### DIFF
--- a/inputs/select.vue
+++ b/inputs/select.vue
@@ -83,7 +83,7 @@
     :label="label"
     :floatingLabel="true"
     :help="args.help"
-    :disabled="disabled"
+    :disabled="isDisabled"
     :error="errorMessage"
     :invalid="isInvalid"
     iconPosition="right"
@@ -114,7 +114,7 @@
     data() {
       return {
         listOptions: [],
-        isDisabled: false
+        userDisabled: false
       };
     },
     mounted() {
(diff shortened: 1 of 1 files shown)


phyllisstein:
Hey @macgyver! Thanks for your patience. I just cut a `v10.0.2` release with your changes. Godspeed.

PR #1547, opened 2022-10-20
Only admins can add new tags

Jess2896:


james-owen (changes requested):
I wonder if we could set this to be a toggleable per kiln or clay instance? Otherwise this would be a breaking change.

PR #1554, opened 2022-12-15
🍕 Fix scheduling/unscheduling workflow

jjpaulino:
The UI when scheduling/rescheduling/unscheduling was not being updated as it should have been because of a few pieces of state that needed to be changed! 

Those things are addressed here, plus the linter decided to fix some files.

Before this, when clicking `unschedule`, the UI wouldn't react and it wouldn't show the UI back to its default state (where the page is not scheduled).

### DIFF
--- a/lib/drawers/publish-layout.vue
+++ b/lib/drawers/publish-layout.vue
@@ -274,6 +274,8 @@
           })
           .then(() => {
             store.commit(FINISH_PROGRESS);
+            // reset date and time values
+            this.clearScheduleForm();
             store.dispatch('showSnackbar', 'Unscheduled Layout');
           });
       },
@@ -324,6 +326,7 @@
       },
       clearScheduleForm() {
         this.dateValue = null;
+        this.timeValue = null;
         this.$refs.timepicker.clear();
       },
       saveTitle() {
--- a/lib/drawers/publish-page.vue
+++ b/lib/drawers/publish-page.vue
@@ -228,7 +228,7 @@
         const store = this.$store;
 
         store.commit(START_PROGRESS);
-        store.dispatch('unschedulePage', this.uri)
+        store.dispatch('unschedulePage')
           .catch((e) => {
             store.commit(FINISH_PROGRESS, 'error');
             log.error(`Error unscheduling page: ${e.message}`, { action: 'unschedulePage' });
@@ -241,6 +241,8 @@
           })
           .then(() => {
             store.commit(FINISH_PROGRESS);
+            // reset date and time values
+            this.clearScheduleForm();
             store.dispatch('showSnackbar', 'Unscheduled Page');
           });
       },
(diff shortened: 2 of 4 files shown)


lib/drawers/publish-page.vue
```diff
@@ -326,6 +328,7 @@
       },
       clearScheduleForm() {
         this.dateValue = null;
+        this.timeValue = null;
```
jjpaulino, line 331: This property needs to be cleared too in order to reflect the changes in the UI.

lib/drawers/publish-page.vue
```diff
@@ -228,7 +228,7 @@
         const store = this.$store;
 
         store.commit(START_PROGRESS);
-        store.dispatch('unschedulePage', this.uri)
+        store.dispatch('unschedulePage')
```
jjpaulino, line 231: This action should not receive a string as a parameter.

lib/drawers/publish-layout.vue
```diff
@@ -324,6 +326,7 @@
       },
       clearScheduleForm() {
         this.dateValue = null;
+        this.timeValue = null;
```
jjpaulino, line 329: This property needs to be cleared too in order to reflect the changes in the UI.

james-owen (approved):

lib/page-data/actions.js
```diff
@@ -220,13 +220,13 @@ export function schedulePage(store, { uri, timestamp }) {
...
+    scheduledId = btoa(pageUri),
     prefix = _.get(store, 'state.site.prefix'),
-    scheduleUri = `${prefix}${scheduleRoute}/${scheduleId}`;
+    scheduledUri = `${prefix}${scheduleRoute}/${scheduledId}`;
 
-  store.dispatch('startProgress', 'scheduled');
+  store.dispatch('startProgress', 'unscheduled');
 
-  return remove(scheduleUri)
+  return remove(scheduledUri)
```
reubenson, line 229: was this the main issue?

reubenson (approved):
nice sleuthing work!!

lib/page-data/actions.js
jjpaulino: The main issues were [here](https://github.com/clay/clay-kiln/pull/1554/files#diff-a05c20d36fd595f8e9f570e888fc5bbde303b1a4d661063d6ddd81dd49e480e9R231) and [here](https://github.com/clay/clay-kiln/pull/1554/files#diff-a05c20d36fd595f8e9f570e888fc5bbde303b1a4d661063d6ddd81dd49e480e9R331).

The first one was causing [this function](https://github.com/clay/clay-kiln/blob/155d71f09e893eea6672258e6ea4d7783aa8a299/lib/page-data/actions.js#L221) to behave completely different because of the incorrect parameter being passed and the second issue was preventing the UI from being fully updated!

PR #1558, opened 2023-02-08
🍕 Fix underline hyperlink text not being displayed while editing paragraph

jjpaulino:
This css line prevents the styling of the clay instance to kick in, making the anchor tags text to not look underlined when editing on a `wysiwyg` input. Removing it allows to pull in any specific/custom styles that exist


james-owen (changes requested):
Looks good. I don't think the dist files should be committed.

james-owen (approved):

Jess2896 (approved):

